PR #22, opened 2021-10-28
[Don't merge] Update to next to `v12.0.2`

lfades:
~Currently failing because `12.0.1` doesn't include this fix: https://github.com/vercel/next.js/pull/30466, we can either wait for `12.0.2` or switch to canary.~


leerob (approved):

lfades:
Blocking the PR because I have found issues in the following two previews while looking for errors:

- https://edge-functions-redirects-upstash-git-update-to-next-latest.vercel.sh/ -> navigations don't work
- https://edge-functions-query-params-filter-git-update-to-next-latest.vercel.sh/ -> redirects don't work

PR #29, opened 2021-11-01
use universal jwt/jose

panva:
`jsonwebtoken` is not a universal library, it does not support the Cloudflare Workers runtime (Web Cryptography API) and when present in a build it'll pull enormous, incomplete, and inefficient node's crypto module polyfills.

This updates the examples that relied on either `jsonwebtoken` or `@tsndr/cloudflare-worker-jwt` to just use `jose`.

A single library that works for developers regardless of the runtime it runs on (currently Browser, Cloudflare Workers, Deno, Electron, Node.js) that uses the native crypto runtime of the platform it runs on and is fully tree-shakeable and conform to the JOSE specifications.

It makes clear distinction w.r.t. key / secret input types. Uint8Array is for symmetric secrets, CryptoKey or KeyObject is for asymmetric key representations such as RSA or EC keys. These can be imported using JWK format (`jose.importJWK()`) or what developers are used to (and can add to their environment variables) PEM `jose.importSPKI()` (public keys), `jose.importX509()` (certificates), `jose.importPKCS8()` (private keys).

When desired it is also possible to use functions that resolve a key, such that key rotation or loading keys asychronously can be done, such as from a public JWKS endpoint using `jose.createRemoteJWKSet` that internally uses again either runtime's native http request capabilities (node's https, or fetch in the web-like runtimes).

Furthermore it is possible to use `jose.EncryptJWT()` / `jose.jwtDecrypt()` to use JWTs in a JWE syntax, meaning they're encrypted, rather than just signed, giving the JWTs a confidentiality attribute.

The library is well documented and has well-maintained types.

### DIFF
--- a/edge-functions/api-rate-limit-and-tokens/lib/api/keys.ts
+++ b/edge-functions/api-rate-limit-and-tokens/lib/api/keys.ts
@@ -1,4 +1,4 @@
-import jwt from '@tsndr/cloudflare-worker-jwt'
+import { jwtVerify } from 'jose'
 import { initRateLimit } from '@lib/rate-limit'
 import { upstashRest } from '@lib/upstash'
 import { API_KEYS, API_KEYS_JWT_SECRET_KEY } from './constants'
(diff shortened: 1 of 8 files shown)


leerob (approved):
This looks like a major improvement 😌

lfades:
@panva Deployment for `api-rate-limit-and-tokens` failed with this:

image

If you build it locally with `next build` it should show you the same build error too

lfades:
`jwt-authentication` is working as expected, not sure if latency had any changes, nothing I can notice.

panva:
I'm going to open a new PR as I deleted the original fork it comes from earlier today and a push to the same branch on a new fork is not getting picked up.

panva:
See #30

PR #30, opened 2021-11-02
use universal jwt/jose

panva:
`jsonwebtoken` is not a universal library, it does not support the Cloudflare Workers runtime (Web Cryptography API) and when present in a build it'll pull enormous, incomplete, and inefficient node's crypto module polyfills.

This updates the examples that relied on either `jsonwebtoken` or `@tsndr/cloudflare-worker-jwt` to just use `jose`.

A single library that works for developers regardless of the runtime it runs on (currently Browser, Cloudflare Workers, Deno, Electron, Node.js) that uses the native crypto runtime of the platform it runs on and is fully tree-shakeable and conform to the JOSE specifications.

It makes clear distinction w.r.t. key / secret input types. Uint8Array is for symmetric secrets, CryptoKey or KeyObject is for asymmetric key representations such as RSA or EC keys. These can be imported using JWK format (`jose.importJWK()`) or what developers are used to (and can add to their environment variables) PEM `jose.importSPKI()` (public keys), `jose.importX509()` (certificates), `jose.importPKCS8()` (private keys).

When desired it is also possible to use functions that resolve a key, such that key rotation or loading keys asychronously can be done, such as from a public JWKS endpoint using `jose.createRemoteJWKSet` that internally uses again either runtime's native http request capabilities (node's https, or fetch in the web-like runtimes).

Furthermore it is possible to use `jose.EncryptJWT()` / `jose.jwtDecrypt()` to use JWTs in a JWE syntax, meaning they're encrypted, rather than just signed, giving the JWTs a confidentiality attribute.

The library is well documented and has well-maintained types.

Replaces #29

### DIFF
--- a/edge-functions/api-rate-limit-and-tokens/lib/api/keys.ts
+++ b/edge-functions/api-rate-limit-and-tokens/lib/api/keys.ts
@@ -36,7 +36,7 @@ export const tokenRateLimit = initRateLimit(async (request) => {
   let payload: ApiTokenPayload
   try {
     const verified = await jwtVerify(token, new TextEncoder().encode(API_KEYS_JWT_SECRET_KEY))
-    payload = <ApiTokenPayload>verified.payload
+    payload = verified.payload as ApiTokenPayload
   } catch (err) {
     return tokenExpired()
   }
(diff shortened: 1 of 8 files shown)


panva:
That's not bad considering there's now support for 6 more JWS algorithms in the bundled code ;)

lfades (approved):
`api-rate-limit-and-tokens` is now working as expected so everything looks right!

@panva Thank you so much for the PR! 💯 🌟

PR #91, opened 2022-01-26
Adds combining data fetching strategies example

goncy:
This example shows how to use different data fetching strategies to avoid switching from static pre-rendering.

image

image


solutions/combining-data-fetching-strategies/components/Snippet.tsx
```diff
@@ -0,0 +1,11 @@
+import type {VFC, ReactChild} from "react";
+
+const Snippet:VFC<{children: ReactChild}> = ({children}) => {
```
gbibeaul, line 3: I learned something today! 

Is that better than React.FC ? I see that VFC does not imply children.

gbibeaul (approved):
🥇 🚀

solutions/combining-data-fetching-strategies/components/Snippet.tsx
goncy: I just realized that this component is not being used, thanks! I think its better in the meaning that you have to be more explicit. It has some other improvements but I don't quite remember them now 🤔

PR #116, opened 2022-01-31
Add reuse responses example

goncy:
This example shows how to use fs to cache responses between getStaticPaths and getStaticProps through calls to reduce bandwidth and processing time.

image


dominiksipowicz (approved):
Great example!

PR #118, opened 2022-02-01
Feature: `breakpoint server` as middleware with user-agent detection and option to ISR to have an improved visual speed to our enduser

maapteh:
I created an example with using middleware to detect the best educated guess for a breakpoint utilizing the User-Agent. And then passing this initial breakpoint to the application. Which then takes it from there and sees for the real breakpoint and updates the UI accordingly. 

Because of middleware it became possible to have a better strategy to be right in the first render-cycle to our clients. So the best UX with minimal reflows!

It also shows how it can build the types of pages we defined, and serve these from the SSG.

```
Page                                       Size     First Load JS
┌   /_app                                  0 B            76.2 kB
├ ƒ /_middleware                           40.8 kB         117 kB
├ ● /[breakpoint] (ISR: 60 Seconds)        510 B          76.7 kB
├   ├ /sm
├   ├ /md
├   └ /lg
└ ○ /404                                   194 B          76.4 kB
+ First Load JS shared by all              76.2 kB
  ├ chunks/framework-91d7f78b5b4003c8.js   42 kB
  ├ chunks/main-a89779f987984dde.js        27.8 kB
  ├ chunks/pages/_app-fb44cafc2a8f476d.js  5.58 kB
  ├ chunks/webpack-592b97d0377c68ad.js     812 B
  └ css/b996d6184d403e5d.css               3.35 kB
```

_More background information can be found at https://mpth.medium.com/render-server-side-for-specific-viewports-with-option-to-cache-1196fbf936c3 and i thought an example explains way more then some loose file references._


maapteh:
@leerob how to get this reviewed?

edge-functions/breakpoint-server/pages/[breakpoint]/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    props: {
+      breakpoint,
+    },
+    revalidate: 60,
+  }
+}
+
+export default function HomePage() {
+  const size = useBreakpoint()
+  const isSmall = size === Breakpoint.xs || size === Breakpoint.sm
```
Comment (lfades) [inline edge-functions/breakpoint-server/pages/[breakpoint]/index.tsx:36]: Am I right in thinking that the breakpoint is actually going to be decided by the `getBreakpoint` function inside `useEffect`? Even though that works, I would like to see it being decided by the edge instead, otherwise it's recommended to use CSS to avoid a CLS of content in the browser.

lfades (changes requested):
Hi @maapteh 👋  thank you for the PR!

I added a small comment, and I'm blocking changes because after it's resolved I'll update versions to match the other examples that were recently updated.

edge-functions/breakpoint-server/pages/[breakpoint]/index.tsx
Comment (maapteh) [reply edge-functions/breakpoint-server/pages/[breakpoint]/index.tsx:36]: But @lfades thats actually the case, see https://github.com/vercel/examples/blob/f1d50db0c420e69d21b7f46db833fe53a31585e1/edge-functions/breakpoint-server/pages/_middleware.ts#L34 in the middleware the initialBreakpoint is passed to the application at https://github.com/vercel/examples/blob/f1d50db0c420e69d21b7f46db833fe53a31585e1/edge-functions/breakpoint-server/pages/_app.tsx#L20 

So server-side its in the breakpoint we got on the edge. Then after that client-side kicks in and updates it to the real value. Since user-agent trick is just to get a better fit.

maapteh:
> Hi @maapteh 👋 thank you for the PR!
> 
> I added a small comment, and I'm blocking changes because after it's resolved I'll update versions to match the other examples that were recently updated.

@lfades what is the status ?

maapteh:
@lfades thank you for unblocking :)

PR #127, opened 2022-02-01
Delete vercel.json for API rate limit

leerob:


vercel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "github": {
+    "silent": true
```
leerob, line 3: Also turning off our comments for now, it's a bit noisy

okbel:
Even if it's noisy, those are the comments we use to see if deployments are working.

lfades (approved):
Re-opened the PR and updated it to only include the change of removing the `vercel.json`, thank you!

PR #87, opened 2022-01-24
Update to all examples.

okbel:
## Update to all examples

Bumps all safe versions in every example/solution. Next on our todo is running an script weekly to automatize this process.  

### Upgrade Tailwind Config to v3 
Tailwind introduces breaking changes in the v3. All examples have the following replacements done.

1. Removing `darkMode` -> `darkMode.*$`
2. Removing the `variants` configuration -> `variants:((.|\n)*)(\}\,)`
3. Removing and/or replacing `purge` and configuring content sources: `purge:((.|\n)*?)(\]\,)`
4. Updating colors -> just one.
5. Updating outline config (none)
6. Removing unneeded classNames because of Automatic transforms and filters

PR #132, opened 2022-02-02
Add update script and fix header path in examples UI

lfades:
Also updated its related deps (postcss, autoprefixer), added a script that can be used to update a dep in all examples , and updated the examples ui package to fix a style issue in the example's path in the header


okbel:
It's already done at #87

goncy (approved):

PR #133, opened 2022-02-02
feat: use plop to generate example template

dominiksipowicz:
## Why?

The purpose of this repository is to provide examples of best practices. As we add more over time, we want to use a consistent look and feel and standardize the process. It also makes it easier for the community to contribute quickly by having a script that generates scaffolding.

## how to use it?

in the root folder run `npm run new-example`

Screenshot 2022-02-02 at 15 26 30

ticket reference: SOL-173
tool: https://www.npmjs.com/package/plop


plop-templates/example/README.md
```diff
@@ -0,0 +1,41 @@
+# -- PLOP TITLE HERE -- example
+
+This example shows how -------------------------
+
+## Demo
+
+https://-----------------------.vercel.app
```
goncy, line 7: Probably we can do this line the same width of the above one so we can find and replace it

plopfile.js
```diff
@@ -0,0 +1,175 @@
...
+        {
+          type: 'modify',
+          path: `{{exampleScopeFolder}}/${plopExampleName}/pages/_app.tsx`,
+          pattern: /(-- PLOP TITLE HERE --)/gi,
+          template: `${data.name}`,
+        },
+        {
+          type: 'modify',
+          path: `{{exampleScopeFolder}}/${plopExampleName}/pages/_app.tsx`,
+          pattern: /(-- PLOP PATH HERE --)/gi,
```
goncy, line 162: Awesome! Probably we can replace this text with something else? Like [[:ploppath:]] ?

plop-templates/example/README.md
```diff
@@ -0,0 +1,41 @@
...
+Deploy the example using [Vercel](https://vercel.com?utm_source=github&utm_medium=readme&utm_campaign=next-example):
+
+[![Deploy with Vercel](https://vercel.com/button)](https://vercel.com/new/git/external?repository-url=https://github.com/vercel/examples/tree/main/-- PLOP PATH HERE --&project-name=-- PLOP EXAMPLE NAME HERE --&repository-name=-- PLOP EXAMPLE NAME HERE --
+
+### Clone and Deploy
+
+Execute [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app) with [npm](https://docs.npmjs.com/cli/init) or [Yarn](https://yarnpkg.com/lang/en/docs/cli/create/) to bootstrap the example:
+
+```bash
+npx create-next-app --example https://github.com/vercel/examples/tree/main/-- PLOP PATH HERE --
```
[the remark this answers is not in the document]
dominiksipowicz: yes this is auto replaced. The "variables" for templates use `-- PLOPVAR HERE --` pattern for replacing in template

plop-templates/example/README.md
dominiksipowicz: yeah, I was thinking about this. But sometimes we have a different URL and when it's auto-populated we will take it for granted and with those huge dashes, it's obvious to update this. But yeah I can automate it. The question is if code reviewer will always check this (before and after merge).

README.md
```diff
@@ -6,13 +6,17 @@
...
 
 ## Examples
 
 - [Edge Functions](/edge-functions) – Edge Functions are currently in Public Beta on Vercel. [Read the docs here.](https://vercel.com/docs/concepts/functions/edge-functions)
 - [Solutions](/solutions) – Demos, Architectures and Best Practices
 
+## Adding new example
+
+Use `npm run new-example` to generate scaffolding for new example.
+
```
lfades, line 19:
```suggestion
### Adding a new example

To quickly start contributing with a new example, run the following commands:

```bash
npm i
npm run new-example
```

plopfile.js
```diff
@@ -0,0 +1,175 @@
+module.exports = function (plop) {
```
lfades, line 1: Plop docs are using ES for their examples, can we use it here too?

plop-templates/example/README.md
goncy: Probably the project won't be created until merge, so this can be modified later in case than needed, if we don't fill it we still have to come back to this, I preffer to come and change a few chars sometimes but don't change it most of the times

dominiksipowicz: done

plopfile.js
dominiksipowicz: done

plop-templates/example/README.md
```diff
@@ -0,0 +1,41 @@
+# -- PLOP TITLE HERE -- example
+
+This example shows how -------------------------
```
goncy, line 3: Can we fill this too? I mean, so we don't have to change anything, that would be awesome!

dominiksipowicz: you mean to create another CLI plop input to ask for a long description when creating the example? So I prepopulate title with `-- PLOP TITLE HERE --` and I can do that but I believe it would make sense for developer to populate this long description after the example is created and those large dashes indicate that this should be manually populated.

lfades (changes requested):
- Tailwind CSS is required (and `pages/_app.tsx` should always be there) because it's used for the layout added by `@vercel/examples-ui`. We want all examples to have a consistent layout.
- `package.json` shouldn't have `private: true` and it's missing `eslint` in dev dependencies (required by `next lint`)
- The generated API route should use `req`/`res` instead of `request`/`response` to match our other educational content
- Optional: The same as above happens for middleware, `req`/`ev` instead of `request`/`event`, the Next.js and Vercel docs are interchanging the two from one example to the other for some reason, I would stitch to the shortest version as that one is more likely to persist in the future
- The `Vercel.json` and `Vercel Serverless Functions: Hello world` options shouldn't be there until we have one use case for them

The generator is awesome 💯

lfades (approved):
Added some minor changes and it looks good to me!

PR #136, opened 2022-02-03
Upstash rewrite example

goncy:
An example on how to rewrite on the edge using upstash:

https://rewrites-upstash-ljdggnhmk-goncy-vercelcom.vercel.app/

image


lfades (changes requested):
`getServerSideProps` is the alternative but it doesn't require using redirects like in the first graph.

The next graph should be using Edge + `getStaticProps` instead of `getServerSideProps`

dominiksipowicz (approved):
LGTM

lfades (approved):
Added some improvements to the writing and LGTM!

Thank you Goncy, great example 💯

PR #156, opened 2022-02-18
Add logging example

goncy:
This example shows how to implement logging at the edge using logtail

image


edge-functions/logging/pages/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          name="description"
+          content="Vercel example on how to use logging on middleware"
+        />
+        <link rel="icon" href="/favicon.ico" />
+      </Head>
+
+      <section className="flex flex-col gap-6">
+        <Text variant="h1">Logging from the edge</Text>
+        <Text>
+          Logging gives us visibility on usage, errors, etc. The edge has plenty information about the request and loggin calls are often non-blocking so this makes the edge a perfect place for implementing it. All we have to do is call our logging service api using fetch from our `_middleware` file:
```
leerob, line 31:
```suggestion
          Logging gives us visibility on usage, errors, etc. The edge has plenty of information about the request and logging calls are often non-blocking, so this makes the Edge a perfect place for implementing logging. All we have to do is call our logging API using fetch from our `_middleware` file:
```

ValentinH:
Thanks for creating this example!
Is there a way to log the response status code? As far as I understand I don't think this is possible as the middleware is only executed before the destination is reached.

goncy:
> Thanks for creating this example! Is there a way to log the response status code? As far as I understand I don't think this is possible as the middleware is only executed before the destination is reached.

Hey @ValentinH! Which response, the log response or the page response? If its the later you will have to do that on the page as the response hadn't completed at this point

ValentinH:
Indeed, I'm referring to the page response. Do you know if there are plans to had a feature to the middlewares after the request has been handled by the page/API handler and before sending it back to the client?

goncy:
> Indeed, I'm referring to the page response. Do you know if there are plans to had a feature to the middlewares after the request has been handled by the page/API handler and before sending it back to the client?

I don't think so

edge-functions/logging/pages/_middleware.ts
```diff
@@ -0,0 +1,23 @@
+import { NextRequest } from 'next/server'
+
+export default function middleware(req: NextRequest) {
+  // Only log for visited pages
+  if (req.page) {
```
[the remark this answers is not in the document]
goncy: I don't see it but got it from the types :thonk:, its the only easy way to verify if the request is for a page 👀

edge-functions/logging/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Demo
+
+https://edge-logging.vercel.app
+
+## How to Use
+
+### Fill environment variables
+
+Rename `.env.example` to `.env` and fill `LOGTAIL_TOKEN` with your [Logtail](https://logtail.com) token.
```
lfades, line 13: `.env.local` is the place for secrets.
 
```suggestion
Rename `.env.example` to `.env.local` and fill `LOGTAIL_TOKEN` with your [Logtail](https://logtail.com) token.
```

edge-functions/logging/README.md
```diff
@@ -0,0 +1,47 @@
+# logging example
+
+This example shows how to implement logging at the edge using logtail.
```
lfades, line 3:
```suggestion
This example shows how to implement logging at the edge using [Logtail](https://betterstack.com/logtail).
```

edge-functions/logging/pages/_middleware.ts
```diff
@@ -0,0 +1,27 @@
+import { NextRequest } from 'next/server'
+
+// Regex for public files
+const PUBLIC_FILE = /\.(.*)$/
+
+export default function middleware(req: NextRequest) {
+  // Only log for visited pages
+  if (!PUBLIC_FILE.test(req.nextUrl.pathname)) {
+    // We just fire and forget this request as we don't want to block the request until completion
```
lfades, line 9:
```suggestion
    // We fire and forget this request to avoid blocking the request until completion
    // and let logging occur in the background
```

edge-functions/logging/pages/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+          our <Code>_middleware</Code> file:
+        </Text>
+        <Snippet>
+          {`// Regex for public files
+const PUBLIC_FILE = /\.(.*)$/
+
+export default function middleware(req) {
+  // Only log for visited pages
+  if (!PUBLIC_FILE.test(req.nextUrl.pathname)) {
+    fetch('https://in.logtail.com', {
```
lfades, line 44: The same explanation comment from the middleware page should be here too

```suggestion
    // We fire and forget this request to avoid blocking the request until completion
    // and let logging occur in the background
    fetch('https://in.logtail.com', {
```

edge-functions/logging/pages/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    })
+  }
+}`}
+        </Snippet>
+        <Image src={board} alt="Middleware logging implementation" />
+        <Text>
+          This way our call will not block the execution (because we do not want
+          to block the request until completion as we do not need the response)
+          while logging the information in our service of choice. This example
+          uses <Link href="https://logtail.com/">Logtail</Link> but you can use
```
lfades, line 68: Links to external sites should open in a new tab

lfades (changes requested):
Added another review 🕵️

edge-functions/logging/README.md
```diff
@@ -0,0 +1,47 @@
+# logging example
```
lfades, line 1:
```suggestion
# Logging at the edge
```

lfades (approved):

PR #161, opened 2022-02-21
Update to latest Next.js in all examples

lfades:
This PR updates all examples to the latest Next.js version, and also updates the rewrites and redirects that are happening inside middleware to always use an absolute URL.


dominiksipowicz:
@lfades how come all the builds are canceled? In this particular case, they all actually should be deployed.

dominiksipowicz (approved):

PR #169, opened 2022-02-23
use prettier in vscode on save

dominiksipowicz:
### Description

This will ensure to run prettier on save. I was missing this on my setup.

PR to PR @lfades ;)

### Best Way to Test

Use VS Code :)

### Type of Change

- [ ] New Example

### New Example Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
+}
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
+}
```
lfades, line 4: What happens if the extension isn't installed?

dominiksipowicz: I think nothing. the setting will be set but not used :)

PR #171, opened 2022-02-24
Fix bel name on codeowners

lfades:


dominiksipowicz (approved):

.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
 # Learn how to add code owners here:
 # https://help.github.com/en/articles/about-code-owners
 
-* @bel @lfades @goncy @gbibeaul @dominiksipowicz
```
gbibeaul, line 4: Should we not use github teams instead?

PR #173, opened 2022-02-24
Implement image fallback example

goncy:
### Adding an example on how to use a fallback image for next/image

image


gbibeaul (approved):
WOW really useful!

PR #178, opened 2022-03-01
Chore(auth-with-ory): Make the example with Ory feel more like the other examples

gbibeaul:
### Description

Adapts the auth with ory example to fit better the vercel example standard:

- remove prettier
- better domain in readme
- change package name

### Type of Change

- [x] Chore/Cleanup
- [ ] New Example

### New Example Checklist


dominiksipowicz (approved):

lfades (commented):
@gbibeaul The deployment doesn't have the environment variables that the example uses, doing the sign in ends up with an error. This was expected because we didn't develop the example to begin with so we'll need to setup those.

We should also clear the index page a bit, like removing the boxes of the getting started template of Next.js

PR #182, opened 2022-03-03
Feat/refactoring

Y0moo:
### Description

- fully replaced `wagmi` by `react-moralis`
- fixed bug allowed to mint NFTs on all chains
- fixed UI bug with uploading image
- improved routing+states(`isAuthenticated`, current Chain and etc.)
- added subdomain in env for getting image links automatically in next.config
- added error message if Rarible Plugin is not installed
- better state handling for "loading" buttons
- added .env.example

### Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature in existing example


gbibeaul (approved):
Solid work! Thanks a lot for the contribution 🚀 🔥

PR #190, opened 2022-03-14
suggar high snippet added

gbibeaul:
### Description

adds a snippet component to the example ui component library. We often copy the snippet component so this will help standardize the snippets. I added [Sugar-High](https://github.com/huozhi/sugar-high) from @huozhi since it's insanely light and fits our needs well.

image

The color scheme is customizable if we want to change anything inside

### Best Way to Test

1. pull locally
2. in any example that uses a snippet `npm i ../../packages/ui `
3. replace the snippet by the one from the lib. Usage is the same.

### Type of Change

- [ ] New Example
- [x] Repo improvement

### New Example Checklist


packages/ui/snippet/snippet.tsx
```diff
@@ -0,0 +1,13 @@
+import s from './snippet.module.css'
+import { highlight } from 'sugar-high'
+
+const Snippet: React.FC = ({ children }) => {
+  const __html = children ? highlight(children?.toString()!) : ''
```
lfades, line 5: Should this be inside `useMemo`? to avoid using `highlight()` again

packages/ui/base.css
```diff
@@ -69,4 +69,12 @@
...
   --shadow-hover: 0 30px 60px rgba(0, 0, 0, 0.12);
   --shadow-large: rgba(0, 0, 0, 0.12) 0px 30px 60px 0px;
+
+  --snippet-class: #2d5e9d;
+  --snippet-identifier: var(--accents-8);
+  --snippet-sign: #8996a3;
+  --snippet-string: var(--geist-cyan-dark);
+  --snippet-keyword: var(--lightblue);
+  --snippet-comment: #8996a3;
+  --snipet-literals: #03066e;
```
lfades, line 79: Can we use the colors of our styleguide: https://vercel.com/design/color

In particular all the colors with `geist`

lfades (changes requested):
Added some comments

lfades (approved):
Added some changes:
- Added custom webpack config to make sure the package uses the react version provided by the app, otherwise hooks don't work, like the one added by `Snippet`
- Added eslint to the package or otherwise the eslint step of commits would fail
- Removed the peer dependency change to the Next.js version (lower is better)

PR #194, opened 2022-03-18
chore: replace deprecated String.prototype.substr()

CommanderRoot:
### Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New Example

### New Example Checklist

### DIFF
--- a/solutions/monorepo/packages/utils/src/index.ts
+++ b/solutions/monorepo/packages/utils/src/index.ts
@@ -13,9 +13,9 @@ export function randomColor() {
  * our demo purposes.
  */
 export function matchingTextColor(color: string) {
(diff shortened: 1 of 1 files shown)


lfades (approved):
Thank you!

PR #198, opened 2022-03-23
Add loading web fonts example

goncy:
This example shows how to load web fonts without compromising performance, its based on https://www.industrialempathy.com/posts/high-performance-web-font-loading/ this article from Malte.

image

Deploy: https://loading-web-fonts-goncy-vercelcom.vercel.app/


solutions/loading-web-fonts/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+  as="font"
+  type="font/woff2"
+  crossorigin
+/>`}</Snippet>
+        <Text className="border-l-4 pl-4">
+          If we use <Code>Google Fonts</Code> or <Code>Typekit</Code>,{' '}
+          <Link href="https://nextjs.org/docs/basic-features/font-optimization">
+            Next.js Font Optimization
+          </Link>{' '}
+          will automatically add the preconnect link.
```
cramforce, line 96: This comes right after talking about preload, but it isn't the same thing. You'd still want to manually add the preload independent of the preconnect if you can.

solutions/loading-web-fonts/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+          shift in the UI.
+        </Text>
+        <Text>
+          One possible solution is having an adjusted fallback font that tries
+          to match the used space of our custom font to reduce the layout shift.
+          You can use{' '}
+          <Link href="https://www.industrialempathy.com/perfect-ish-font-fallback/?font=Inter">
+            this tool
+          </Link>{' '}
+          to find which fallback font might match your custom font.
```
cramforce, line 140: You should mention somewhere that this only works well if `line-height` is set to a non-default value.

solutions/loading-web-fonts/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+          convallis mauris dui, non placerat urna fringilla at. Suspendisse a
+          orci quis arcu tristique sagittis at sed leo. Nam quis neque dapibus,
+          semper nisl non, lacinia ligula. Duis viverra a nisl ut consectetur.
+          Ut elementum fringilla odio viverra egestas. Morbi aliquet lorem
+          lorem, in suscipit nibh tempus quis. Mauris gravida dapibus odio,
+          vitae interdum lectus pretium ut. Vivamus tincidunt laoreet
+          pellentesque. Praesent tincidunt elementum tempus.
+        </FontMixer>
+        <Text>
+          Fonts with different sizes not only may create a layout shift in its
```
cramforce, line 129: Layout shift only happens with `font-display` values other than `optional`. I think it still makes sense to recommend this technique because it makes the layouts more similar which may avoid layout bugs.

solutions/loading-web-fonts/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+            text.
+          </li>
+          <li>
+            <Image src={displaySwap} alt="font-display: swap" />
+            <Code>font-display: swap</Code>: Will display the text using the
+            fallback font until our custom font is loaded and will then swap the
+            fallback font with the custom font.
+          </li>
+          <li>
+            <Image src={displayOptional} alt="font-display: optional" />
```
cramforce, line 183: There is the nuance that if you do not have a preload for the font (which triggers a special heuristic) then you will never see the font on first page load in a session, even if it is in disk cache (you'd only see it if it is in memory cache)

gbibeaul (approved):
Tiny proposal to change Malte Ubl's name to a link to some of his profiles

solutions/loading-web-fonts/globals.css
```diff
@@ -0,0 +1,5 @@
+@font-face {
```
cramforce, line 1: Wouldn't it be awesome to also use this technique on vercel.com? Looks like we currently are not.

cramforce (dismissed):

Issue #119, opened 2022-02-01
'logging' example is not provided; no pattern provided


sgharms:
Hello,

The https://nextjs.org/docs/middleware#examples page suggests that there are number of examples of various types in this repo

```
Authentication
Bot protection
Redirects and rewrites
Handling unsupported browsers
Feature flags and A/B tests
Server-side analytics
Advanced i18n routing requirements
Logging
```

all point to the same resource: 

https://github.com/vercel/examples/tree/main/edge-functions

Keying in on **Logging**, searching the Readme for the string `oggi` yields no match. Similarly `ack` through the repo for `/oggi/i` and you will only find a mention in `feature-flag-configcat/` example. Furthermore, this is logging from a _client_ function not from a server function e.g. the middleware function.

Absent any directions, I ahem, duplicated this error https://github.com/vercel/next.js/discussions/13214

Further research at https://nextjs.org/docs/api-reference/edge-runtime#unsupported-apis, a leap whose connection was made my reading URL paths, suggests this is not possible. OK, I can live with that, but then it becomes all the more important that it be documented how to log server-side data in the middlewares. 

I suspect most people are implementing a custom server to get `'fs'` access, a practice that seems discouraged in the docs.

So is there either:

1. An example of doing server-side logging in the Next Middleware (if not...)
2. The expectation that that should be possible and, if so, how will that be possible if `fs` and other core modules are not available (e.g. `process`, `util`, `os`, `buffer`, et al.)

oswaldoacauan:
I had a very frustrating evening yesterday trying to find a simple logging middleware example. The docs pointing nowhere was not very helpful also.

I see this as a basic feature for a framework that is to be production ready same goes for Telemetry which there is no examples to be found.

oswaldoacauan:
@okbel thanks for the update 🙏🏻 . Is there any estimated date for the logging, telemetry and instrumentation examples?

leerob:
https://github.com/vercel/examples/pull/156

sgharms:
I think @leerob has offered something formidable, and it is definitely an improvement that should be included; however, for simplicity’s sake and to establish a general pattern, I’d like to see something where the log file is written on the same machine as the next instance.

[...]

PR #200, opened 2022-03-23
Remove analytics/logging examples

lfades:
### Description

The use case of Analytics and Logging will be better handled in a different way in the future.


goncy (approved):

jasonwilliams:
> ### Description
> The use case of Analytics and Logging will be better handled in a different way in the future.

Hey @lfades can you elaborate what new way will better handle logging?
I was just about to look into middleware to handle this but I've seen you've removed the documentation.

For context we would like to log things about the user from the request but that's not easy to do right now so looked into middleware as a place for that.

lfades:
@jasonwilliams We'll be adding more built-in ways to do telemetry in the future, also for more context: https://github.com/vercel/examples/issues/119#issuecomment-1079769445

PR #205, opened 2022-03-28
Update `.gitignore` in all examples to not ignore `.env`

lfades:
### Description

`.env` is okay to be on a repository, it's `.env*.local` the ones that have to be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ yarn-debug.log*
 yarn-error.log*
 
 # Local ENV files
-.env
 .env.local
 .env.development.local
 .env.test.local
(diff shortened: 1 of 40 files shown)


jaredpalmer:
i suggest keeping .env in gitignore and adding replacing .env.example with the shape of the needed .env

lfades:
@jaredpalmer According to the Next.js docs (https://nextjs.org/docs/basic-features/environment-variables#default-environment-variables) `.env`, `.env.production`, `.env.development` are there to save defaults, or public secrets, their `.local` variants are for secrets that should be ignored by `.gitignore`, the change is to make sure we follow that for `.env`.

> replacing .env.example with the shape of the needed .env

Not sure I understand this part, `.env.example` is used when we want to provide a starting point to creating your `.env.local` with your own secrets, there are no defaults, and if the example requires defaults then we use `.env` and push it to the repo.

dominiksipowicz (approved):

PR #230, opened 2022-04-18
Add "Build Output API" examples

TooTallNate:
### Description

Adds a handful of examples that are "prebuilt" projects that implement the **Build Output API**.

We will be able to link to some of these examples from the relevant sections in the documentation for readers to get more explicit examples of how certain features work.

https://github.com/vercel/examples/blob/add/build-output-api-examples/build-output-api/README.md

### Best Way to Test

Run `vc deploy --prebuilt` on any of the example directories added in this PR to deploy.

### Type of Change

- [x] New Example

### New Example Checklist


EndangeredMassa (approved):
@TooTallNate and I discussed whether or not the `.vercel/output/config.json` file should be required. It seems like there's no real downside and the explicit mention of the file and the version can be useful going forward. It's also technically required inside `static-build` right now. So, we want to make it required.

That means we'll want all of these examples to have that file as well.

EndangeredMassa (approved):

lfades (approved):
- Updated the singular examples to plural (Edge Functions, Serverless Functions) as I think we might eventually include more than 1 function in the example
- Improved the lint-staged config to exclude `.vercel` , same for prettier


@TooTallNate Should we also include an example on how to generate the build output? 🤔

PR #241, opened 2022-04-26
Add more Build Output API examples

TooTallNate:
Adds more examples that we can link to from the docs page.


EndangeredMassa (approved):
These looks awesome. I think we can merge as-is, but I also suggested some tweaks here: https://github.com/vercel/examples/pull/242

EndangeredMassa (approved):

PR #261, opened 2022-05-20
Remove Edge Functions example that writes to the response body

SukkaW:
### Description

Ref: https://github.com/vercel/next.js/pull/36835

Since Next.js v12.1.7-canary.10, Next.js page middleware removes the support of modifying the response body, and here I quote:

> 1. during development, a warning on the console tells developers when they are returning a response (either with `Response` or `NextResponse`).
> 2. at build time, this warning becomes an error.
> 3. at run time, returning a response body will trigger a 500 HTTP error with a JSON payload containing the detailed error.

And here is reasons why the example should be removed:

**edge-functions/cors**

The example uses `new Response(null, { statusCode: 204 })` to respond a CORS preflight request (An `OPTIONS` request), which in Next.js it now means the original body will be returned. It is now impossible to return an empty response from page middleware.

**edge-functions/json-response** and **edge-functions/crypto**

In this example's middleware, it writes a `JSON.stringify`-ed to a returned response body which is now unsupported.

**edge-functions/image-response**

In this example's middleware, it writes a blob (containing an image binary) to a returned response body which is now unsupported.

---

Also, there are still many examples that write to the response body in the middleware, which I will change in a coming PR:

**edge-functions/basic-auth-password**

In this example's middleware, it writes `Auth required` to the returned 401 response. I will change it to `NextResponse.rewrite('/auth-required')` in the coming PR.

**edge-functions/geo-location-country-block**

I will change it to `NextResponse.rewrite` in the coming PR.

### Best Way to Test

N/A

### Type of Change

- [ ] New Example

### New Example Checklist


goncy:
Hi @SukkaW, thanks for realizing that this weren't relevant anymore with the last update, is really useful! I will talk with the team to see if we can keep any of them doing changes like a redirect or a rewrite but most seem safe to delete.

Will let you know soon and thanks again 🙌

feugy:
Indeed, thanks for the hard work @SukkaW!

SukkaW:
> Hi @SukkaW, thanks for realizing that this weren't relevant anymore with the last update, its really useful! I will talk with the team to see if we can keep any of them doing changes like a redirect or a rewrite but most seem safe to delete.
> 
> Will let you know soon and thanks again 🙌

@goncy Thanks for your review!

IMHO these examples shouldn't exist in the first place: those are pages' middleware (where pages should be rendered in the browser) and it doesn't make sense to let a page's middleware return a JSON.

However, the API route of Next.js should be able to switch to edge function runtime instead. I will raise a Feature Request toward Next.js.

lfades (changes requested):
@SukkaW Hi there and thank you for the PR 👋 

We don't want to remove any examples yet, instead we want to update them to use redirects/rewrites, even if that doesn't make a lot of sense, for the case of `crypto, `image-response`, and `json-response` lets update the readme and link to the new API where we take the decision and reasoning behind it, we can remove them if needed some months after the new API is available. The examples have been existing for a while now and removing them breaks shared links.

SukkaW:
> @SukkaW Hi there and thank you for the PR 👋
> 
> We don't want to remove any examples yet, instead we want to update them to use redirects/rewrites, even if that doesn't make a lot of sense, for the case of `crypto, `image-response`, and `json-response` lets update the readme and link to the new API where we take the decision and reasoning behind it. The examples have been existing for a while now and removing them breaks shared links.

Sounds fair.
I also notice that Next.js is adding Edge runtime support for the API route (https://github.com/vercel/next.js/pull/36947). I will change the example to use API route + Edge Runtime when the feature landed.

lfades:
Closing this PR in favor of multiple PRs the team has been making to update the examples in a different way, thank you for the work you did here!

PR #262, opened 2022-05-20
docs: fix README ory example

vinckr:
I noticed there is a small error with the env variables and repo name in the Ory one-click deploy button. 
this changes it to the correct env variable and fixes the name.


vinckr:
found another one, broken/updated link

lfades (approved):
Thank you! 💯

PR #277, opened 2022-06-02
Update rewrites-upstash example

goncy:
Update rewrites-upstash example for middleware GA

### DIFF
--- a/edge-functions/rewrites-upstash/package.json
+++ b/edge-functions/rewrites-upstash/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@vercel/examples-ui": "^0.3.2",
-    "next": "^12.1.0",
+    "next": "^12.1.7-canary.24",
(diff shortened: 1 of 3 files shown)


edge-functions/rewrites-upstash/package.json
```diff
@@ -10,14 +10,14 @@
...
+    "react-dom": "latest"
   },
   "devDependencies": {
-    "@types/node": "^17.0.14",
-    "@types/react": "^17.0.38",
+    "@types/node": "^17",
+    "@types/react": "^18",
     "autoprefixer": "^10.4.2",
     "eslint": "^8.9.0",
     "eslint-config-next": "^12.1.0",
```
lfades, line 23: Setting this to `canary` should avoid an issue with the lint rule for `next/server` that should be happening right now.

```suggestion
    "eslint-config-next": "canary",
```

PR #284, opened 2022-06-06
update json-response middleware

dominiksipowicz:
### Description

- Update JSON response middleware example for middleware GA
- I used a rewrite to serve static file

### Best Way to Test

preview link


edge-functions/json-response/pages/index.tsx
```diff
@@ -0,0 +1,3 @@
+const Page = () => <div>lol</div>
+
+export default Page
```
lfades, line 3: @dominiksipowicz Let's remove this, the example should show a JSON in the index page just as it did before.

lfades (changes requested):

edge-functions/json-response/middleware.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @next/next/no-server-import-in-page */
+import { NextResponse, NextRequest } from 'next/server'
+
+export const config = {
+  matcher: '',
```
goncy, line 5: Shouldn't this be '/file' without doing the check below?

dominiksipowicz: the logic should be to run on all pages apart from the file so as to filter downstream. It should be actually `matcher: '/'`. Cheers!

edge-functions/json-response/middleware.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function middleware(req: NextRequest) {
+  if (req.nextUrl.pathname == '/file') {
+    return
+  } else {
+    const url = req.nextUrl.clone()
```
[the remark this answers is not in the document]
dominiksipowicz: cheers!

edge-functions/json-response/middleware.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @next/next/no-server-import-in-page */
+import { NextResponse, NextRequest } from 'next/server'
+
+export const config = {
+  matcher: '/',
+}
+
+export function middleware(req: NextRequest) {
+  if (req.nextUrl.pathname == '/file.json') {
+    return
```
lfades, line 10: This check will never happen.

edge-functions/json-response/middleware.ts
```diff
@@ -0,0 +1,16 @@
...
+  } else {
+    const url = req.nextUrl
+    url.pathname = '/file.json'
+    return NextResponse.rewrite(url)
```
lfades, line 14: Lets use an API route instead of a public file, soon after releasing middleware we'll release a way for API routes to be on Middleware and in that case this middleware will be removed in favor of the API route.

edge-functions/json-response/middleware.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @next/next/no-server-import-in-page */
```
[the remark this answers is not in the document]
dominiksipowicz: done

edge-functions/json-response/middleware.ts
dominiksipowicz: Great feedback, thanks! done

edge-functions/json-response/middleware.ts
dominiksipowicz: done

lfades (approved):

PR #294, opened 2022-06-14
Middleware with  sveltekit

GuiBibeau:
WIP:

- Vercel middleware with sveltekit
- UI lib adapted to sveltekit


packages/ui-svelte/src/Code.svelte
```diff
@@ -0,0 +1,21 @@
+<script>
+  import clsx from 'clsx'
+  export let className = ''
+</script>
+
+<code class={clsx('root', className)}>
```
Rich-Harris, line 6: I'm torn about this. I understand _why_ there's a desire to create this sort of wrapper component (it's a very common mindset in the JS-centric React community), but it's fairly unidiomatic — it's not considered a best practice in Svelte. It's much more common to just use `<code>` rather than `<Code>`, if all the component is doing is adding some styles that could live in a `base.css`.

Firstly, it's much more efficient, both in terms of authoring (no need to import the component, just write `<code>`) and app performance (no need to create potentially hundreds of extra components, no need to run `clsx(...)` for each one, no need to include reactivity code to anticipate those classnames changing when they probably won't).

Secondly, it works with content in HTML, such as you might get from a markdown document or a CMS.

Thirdly, because of how style encapsulation is designed to work, these will behave differently:

```svelte
<code class="extra">here is my code</code>

<style>
  .extra {
    color: hotpink;
  }
</style>
```

```svelte
<script>
  import { Code } from '@vercel/examples-ui-svelte';
</script>

<Code className="extra">here is my code</Code>

<style>
  .extra {
    color: hotpink;
  }
</style>
```

The second example won't work. (This is a frequent subject of debate, but it's deliberate and unlikely to change any time soon!)

So to be honest I think it would be better if this component (and `<Link>`, `<Snippet>` and `<Text>`) didn't exist.

packages/ui-svelte/src/Layout.svelte
```diff
@@ -0,0 +1,37 @@
...
+
+  export let title = 'Vercel Examples'
+  export let description = 'Lear how to use Vercel Middleware in Sveltekit'
+  export let path = ''
+</script>
+
+<svelte:head>
+  <title>{title} - Vercel Examples</title>
+  <meta name="description" content={description} />
+  <link rel="icon" href="/favicon.ico" />
```
Rich-Harris, line 14: This should probably just live in `src/app.html`

packages/ui-svelte/src/Layout.svelte
```diff
@@ -0,0 +1,37 @@
+<script>
+  import './app.css'
+  import Nav from './Nav.svelte'
+  import Vercel from './Vercel.svelte'
+
+  export let title = 'Vercel Examples'
+  export let description = 'Lear how to use Vercel Middleware in Sveltekit'
```
Rich-Harris, line 7: If this component library is designed to be used by multiple examples, then there probably shouldn't be a default value. Though if we do keep it, it's `SvelteKit` rather than `Sveltekit`

```suggestion
  export let description = 'Learn how to use Vercel Middleware in SvelteKit'
```

packages/ui-svelte/src/Nav.svelte
```diff
@@ -0,0 +1,88 @@
+<script>
+  import clsx from 'clsx'
+  import Link from './Link.svelte'
+  export let path = ''
+  const displayPath = ['Vercel Examples']
```
Rich-Harris, line 5: This should be made reactive

```suggestion
  $: displayPath = ['Vercel Examples']
```

packages/ui-svelte/src/Text.svelte
```diff
@@ -0,0 +1,89 @@
...
+    @apply text-lg;
+  }
+
+  .smallText {
+    letter-spacing: 0.1rem;
+    margin-top: 0;
+    padding-left: 0.2rem;
+    @apply uppercase text-blue-500 font-semibold text-sm block mb-4;
+  }
+</style>
```
Rich-Harris, line 89: As mentioned above, I think we should get rid of this component entirely. I don't think there's ever a reason to do

```svelte
<script>
  import { Text } from '@vercel/examples-ui-svelte';
</script>

<Text variant="h1">title</Text>
```

instead of

```svelte
<h1>title</h1>
```

— it's more work to write, it's less efficient, and code written in that style is much less easy to scan

packages/ui-svelte/src/Vercel.svelte
```diff
@@ -0,0 +1,38 @@
+<script>
+  import clsx from 'clsx'
+  export let className = ''
+  export let alt = 'Vercel.com Logo'
```
Rich-Harris, line 4: Does this need to be a prop? When would it ever be anything other than `Vercel.com logo`?

packages/ui-svelte/src/Vercel.svelte
```diff
@@ -0,0 +1,38 @@
...
+<svg
+  width="89"
+  height="20"
+  viewBox="0 0 89 20"
+  fill="none"
+  {alt}
+  class={clsx(className)}
+  xmlns="http://www.w3.org/2000/svg"
+>
+  <path d="M11.5625 0L23.125 20H0L11.5625 0Z" fill="currentColor" />
```
Rich-Harris, line 16: I believe the brand guidelines only allow white or black here — I wonder if the component should reflect that, rather than using `currentColor`?

packages/ui-svelte/tailwind.js
```diff
@@ -0,0 +1,26 @@
+module.exports = {
+    // `content` is replaced instead of extended, so this line has to be added in
+    // the `content` of each app' tailwind.config.js
+    content: ['node_modules/@vercel/examples-ui/ui/**/*.{js,ts,jsx,tsx}'],
```
Rich-Harris, line 4: This feels rather brittle, and a poor user experience. Does the component library really need to use Tailwind?

edge-functions/sveltekit-redirect/package.json
```diff
@@ -0,0 +1,39 @@
+{
+	"name": "sveltekit-redirect",
+	"version": "0.0.1",
+	"scripts": {
+		"dev": "svelte-kit dev",
+		"build": "svelte-kit build",
```
Rich-Harris, line 6: Should this be `vc dev` and `vc build`?

Right now middleware just doesn't work in development; `svelte-kit dev` doesn't understand Vercel middleware. That may change in future (via an expanded adapter API), but in the meantime I'm hoping that `vc dev` will allow it to Just Work in a future version. We should check though, because otherwise this example is DOA :)

edge-functions/sveltekit-redirect/svelte.config.js
```diff
@@ -0,0 +1,19 @@
...
+	// Consult https://github.com/sveltejs/svelte-preprocess
+	// for more information about preprocessors
+	preprocess: [
+		preprocess({
+			postcss: true
+		})
+	],
+
+	kit: {
+		adapter: adapter()
```
Rich-Harris, line 15: I think we probably need to use `adapter-vercel` instead of `adapter-auto`, so that we can set `edge: true`. For now we'd also need to specify `ENABLE_VC_BUILD=1`, no?

Rich-Harris (changes requested):
Thanks! I left more detailed comments inline, but a few high-level thoughts:

* Have you succeeded in deploying this? It doesn't work in development (because `svelte-kit dev` has no way to understand Vercel middleware), and I wouldn't expect it to work in production using `adapter-auto`
* For an example that aims to demonstrate a single feature, there's quite a lot going on. I realise this seems to be the norm for the examples in this repo, but if I'm trying to understand how to add middleware to a project then all the Tailwind/PostCSS/eslint/TypeScript/whatever stuff gets in the way of my understanding. Have we considered making the examples radically simpler?
* I wonder if there's a way to provide an example of using middleware to achieve something that isn't possible in SvelteKit directly? Normally I'd implement this redirect in the `/` route directly (or in `handle` if I wanted a global redirect) rather than putting middleware in front of my entire app. For example doing something like [this](https://www.fastly.com/blog/taming-third-parties-with-a-single-origin-website), or using middleware to enable cross-origin isolation on a prerendered page (we do this on https://learn.svelte.dev so that webcontainers work), or even something cheeky and contrived like rewriting `/cat.jpg` to `/dog.jpg` if a particular cookie is set.
* SvelteKit has a `package` command which is designed for creating component libraries. It's not fully baked yet, but I think it would be much better to use it for `ui-svelte`

packages/ui-svelte/src/Layout.svelte
Rich-Harris: (Also, `favicon.ico` doesn't exist — it's `favicon.png`)

GuiBibeau:
After the reviews, I'll uadate with the following:

- simplified approach in UI lib
- simplify dependencies and what gets in the way of understanding the example


I'll place this as WIP/Draft as it's a big update

lfades:
@Rich-Harris About this:

> For an example that aims to demonstrate a single feature, there's quite a lot going on. I realise this seems to be the norm for the examples in this repo, but if I'm trying to understand how to add middleware to a project then all the Tailwind/PostCSS/eslint/TypeScript/whatever stuff gets in the way of my understanding. Have we considered making the examples radically simpler?

We have TypeScript and eslint for most examples here, those two are fairly straightforward to setup up in Next.js. TS ends up being very similar to JS in React workflows and when using Next.js APIs, it also helps documenting the example better, but I agree it's not required here as it basically doesn't get used.

On Tailwind/PostCSS: it's completely optional. Most examples use Tailwind for styling and so does the examples UI package for React, this actually helps us reducing the code base of examples while still being able to keep them pretty and with a similar UI that includes a header with a deploy button and link to this repo, buttons, code snippet support, etc. I think for Svelte something similar would be needed and shouldn't increase complexity if looking at the code of the UI package shouldn't be required to understand the example.

PR #301, opened 2022-06-16
update api-rate-limit middleware example

dominiksipowicz:
### Description

update to use new middleware

### Demo URL

preview

### Type of Change


edge-functions/api-rate-limit/middleware.ts
```diff
@@ -0,0 +1,14 @@
...
+import { ipRateLimit } from '@lib/ip-rate-limit'
+
+// run only on API routes
+export const config = {
+  matcher: '/api',
+}
+
+export async function middleware(req: NextRequest) {
+  const res = await ipRateLimit(req)
+  if (res.status !== 200) return res
```
lfades, line 11: This won't work if you're rate limited because in that case the method sends a response that's not allowed by the new middleware.

PR #305, opened 2022-06-17
refactor: update image middleware example to use new middleware

dominiksipowicz:
refactor: update image middleware example to use new middleware


edge-functions/image-response/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
leerob, line 2: Interesting - should this be removed?

dominiksipowicz: Yes, this file has been automatically updated by Next.js, so it is safe to delete that line. We removed the line from all recent example changes so it seems consistent.

lfades (changes requested):
The readme also has to be updated.

lfades (approved):

PR #311, opened 2022-06-22
Update middleware in bot-protection-botd

lfades:
### Description

- This is a follow up for https://github.com/vercel/examples/pull/304 to add `matcher` to the middleware as in the Datadome example and simplify it.
- I also updated the root package.json to use the canary version of eslint so that when it executes it uses the same version used by examples

### Type of Change

- [ ] New Example


edge-functions/bot-protection-botd/lib/botd/index.ts
```diff
@@ -33,6 +34,7 @@ export async function botdEdge(
 
   const { pathname } = req.nextUrl
 
+  // Light bot detection is not required for static files or for favicon.ico
   if (STATIC_REGEX_EXCLUSION.test(pathname) && !isFavicon(req)) return
```
goncy, line 38: If the matcher is using `/` and `/blocked` I think it won't execute for static files right?

lfades: It won't, but I don't think we need to remove this check, might be helpful for people implementing it in their own middleware.

PR #313, opened 2022-06-22
Add ISR Example for API Routes

EndangeredMassa:
### Description

Add Build Output API example that uses ISR with API Routes.

### Demo URL

https://isr-revalidation-endangeredmassa.vercel.app/

### Type of Change

- [x] New Example


build-output-api/isr-api-routes/.vercel/output/functions/data.prerender-config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "expiration": 180,
+  "group": 2,
+  "bypassToken": "87734ad8259d67c3c11747d3e4e112d0",
+  "allowQuery": []
+}
```
styfle, line 6: Should we add a newline here for consistency?

build-output-api/isr-api-routes/.vercel/output/functions/revalidate.func/index.js
```diff
@@ -0,0 +1,71 @@
...
+// DO NOT allow unauthenticated users to trigger revalidation: a malicious user could trigger boundless revalidations.
+const authToken = 'a13f94f6-a441-47ca-95fc-9f44f3450295';
+
+function revalidate(host, path) {
+  return new Promise(function (resolve, reject) {
+    const options = {
+      hostname: host,
+      port: 443,
+      path,
+      method: 'GET', // MUST be "GET"; a "HEAD" or "POST" request will not work
```
styfle, line 18:
```suggestion
      method: 'GET', // MUST be "GET" ("HEAD" and "POST" methods will not work)
```

ijjk (commented):
I'm not sure I understand the difference in what's being demonstrated here from the existing demo https://github.com/vercel/examples/tree/main/build-output-api/on-demand-isr 🤔

It seems like this could potentially just be a comment in the other example saying non-HTML data can also be returned for a prerender

build-output-api/isr-api-routes/.vercel/output/functions/revalidate.func/index.js
```diff
@@ -0,0 +1,71 @@
...
+// The bypass token should be a build-time randomly generated string of at least 32 characters.
+// This value should only be exposed in the function config and inside of the function itself.
+// DO NOT expose this value on the client-side: a malicious user could trigger boundless revalidations.
+const bypassToken = '87734ad8259d67c3c11747d3e4e112d0'
+
+// This authToken is a trivial example of authentication, but any strong authentication method will work.
+// DO NOT allow unauthenticated users to trigger revalidation: a malicious user could trigger boundless revalidations.
+const authToken = 'a13f94f6-a441-47ca-95fc-9f44f3450295';
+
+function revalidate(host, path) {
```
louisgv, line 12: I assume this can be done in a nextjs API route as well right? (as long as the token and validation mechanism are properly secured). So basically I can force revalidate the cache with the `x-prerender-revalidate`, and that this should only be done by authorized entity

build-output-api/isr-api-routes/.vercel/output/functions/index.func/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+  // NOTE: the below revalidate link is meant for demo purposes only
+  // actual authentication should be used in actual projects
+  res.end(`
+    <h1>On-Demand Incremental Static Regeneration (ISR) with API Routes Example</h1>
+    <p>The "API Result" below will only get updated after you click the Revalidate link.</p>
+
+    <p>API Result: <span id="result"></span></p>
+
+    <p><a href="/revalidate?authToken=a13f94f6-a441-47ca-95fc-9f44f3450295">Revalidate</a></p>
```
lfades, line 16: Do we want to add a comment here saying that a real application wouldn't send the auth token like this?

lfades (commented):
The PR looks good to me. I suggest adding more details in the How it Works section of the readme because there are many files and it would need careful reading for developers getting started with the behavior of our API or ISR. Providing a guide of what happens when you visit the deployment and click on "Revalidate" would be huge (with text and links to files).

ijjk:
Just wanting to reiterate this comment https://github.com/vercel/examples/pull/313#pullrequestreview-1015702327 as I think having this separate example could cause more confusion than isolating this to the existing example would since users may think it takes different steps when returning content for HTML than for JSON which isn't the case.

build-output-api/isr-api-routes/.vercel/output/functions/revalidate.func/index.js
ijjk: API routes in Next.js do not currently leverage our `Prerender` output type, to leverage `stale-while-revalidate` behavior the `cache-control` header must be set manually so do not currently support On-Demand Revalidate

louisgv: Would it work if the API route set the right cache header?

build-output-api/on-demand-isr/README.md
```diff
@@ -8,10 +8,39 @@ This Prebuilt Deployment example demonstrates how to implement "On-Demand Increm
 
 https://build-output-api-isr.vercel.sh
```
EndangeredMassa, line 9: I'll deploy this on merge of this PR.

EndangeredMassa:
I removed the new example and added a function to the existing ISR example to demonstrate ISR with JSON content.

I did this because it was specifically requested to show in clear terms that this is not just an HTML feature. I think a comment wouldn't have made that as clear, but I definitely agree that an entire separate example was too much.

What do you think of this form? @ijjk @lfades @styfle @TooTallNate

ijjk (approved):

PR #318, opened 2022-06-23
Update crypto example

okbel:
### Description

This new example shows how to use the WebCryptoAPI on both Vercel Edge Functions and Edge Middleware.

### Demo URL
Preview URL

### Type of Change

- [ ] New Example


edge-functions/crypto/next-env.d.ts
```diff
@@ -1,5 +1,5 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
+/// <reference types="next/dist/styled-jsx-types/global" />
```
lfades: Was this a manual change? Next.js should be removing this line entirely 🤔

edge-functions/crypto/package.json
```diff
@@ -4,21 +4,21 @@
...
     "build": "next build",
-    "start": "next start",
-    "lint": "next lint"
+    "dev": "next dev",
+    "lint": "next lint",
+    "start": "next start"
   },
   "dependencies": {
-    "next": "^12.1.0",
+    "next": "12.1.7-canary.45",
```
lfades, line 13: Let's use `canary` here, as those versions change quite fast and we want every example to always be on the latest canary if there are new improvements on the way.

```suggestion
    "next": "canary",
```

edge-functions/crypto/pages/api/crypto.ts
```diff
@@ -33,7 +41,12 @@ export async function middleware(req: NextRequest) {
...
       iv,
+      fromMiddleware,
     }),
     { headers: { 'Content-Type': 'application/json' } }
   )
 }
+
+export const config = {
+  runtime: 'experimental-edge',
+}
```
lfades, line 52: As in middleware.ts, makes more sense to have it in the top

edge-functions/crypto/README.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: Using Crypto in Edge Functions
+name: Using Crypto in Edge Functions and Middleware
```
lfades, line 2: This feels redundant because Middleware is an edge function.

lfades (changes requested):
Added some comments

edge-functions/crypto/package.json
okbel: It doesn't work with canary. It points to another version. Has to be a fixed version for now.

edge-functions/crypto/README.md
okbel: It's not redundant. There's a difference between our Edge Middleware and our Experimental Edge functions (shown and distinguished in the docs)

edge-functions/crypto/next-env.d.ts
balazsorban44: See https://github.com/vercel/next.js/pull/37902

lfades (commented):
@okbel the PR is now including more changes than expected for some reason 🤔

okbel:
@lfades 
Ready to be checked! Pulled from another branch by mistake. Should be good to go now

edge-functions/crypto/next-env.d.ts
lfades: After https://github.com/vercel/next.js/pull/37964 it can be removed now.

lfades (approved):

PR #329, opened 2022-06-24
Update rate limit example for GA

goncy:
Update example for middleware GA


edge-functions/api-rate-limit/middleware.ts
```diff
@@ -0,0 +1,20 @@
...
+  const rateLimitResponse = await ipRateLimit(req)
+  const res = NextResponse.next()
+
+  // Add the rate limit headers to the response.
+  for (const [key, value] of rateLimitResponse.headers.entries()) {
+    res.headers.set(key, value)
+  }
+
+  return res
+}
```
lfades, line 20: This will break once the user is rate limited as that's one of the tasks of the `ipRateLimit` function.

lfades (changes requested):
The example should no longer use `middleware.ts` but move rate limiting to `/api`.

lfades (approved):

PR #330, opened 2022-06-24
Update Feature Flag Configcat

okbel:
### Description
This PR adds updates to the Middleware API

### Demo URL
Preview URL

### Type of Change

- [ ] New Example


edge-functions/feature-flag-configcat/middleware.ts
```diff
@@ -0,0 +1,76 @@
...
+        res.cookies.set(COOKIE_NAME, cookie)
+      }
+
+      return res
+    },
+  },
+]
+
+export const config = {
+  matcher: middlewares.map(({ path }) => path), // This line assumes the path are strings and correclty set.
```
lfades, line 63: Next.js gets this value statically so the `matcher` here shouldn't be working.

lfades (changes requested):
Blocking based on my comment, I'll do follow up updates.

lfades (approved):

PR #332, opened 2022-06-25
Add examples for Edge API Routes.

leerob:
And moved the hello world one from the edge functions folder.


edge-api-routes/cache-control/api/edge.ts
```diff
@@ -0,0 +1,18 @@
+export const config = {
+  runtime: 'experimental-edge',
+}
+
+export default async function handler(req) {
+  return new Response(
```
cramforce, line 6: here and elsewhere change to 
```return Response.json(
    {
      message: 'Hello, world!',
    },
    {
      status: 200,
      headers: {
        'cache-control': 'public, s-maxage=1200, stale-while-revalidate=600',
      },
    }
  )
  ```

leerob: We'll have to swing back on this:

> Response.json can’t be added right now because `.json` is autogenerated and can't be augmented.

cramforce: Too bad, because it absolutely works, just doesn't compile :)

lfades (approved):

PR #338, opened 2022-06-27
Upgrade Feature Flag Posthog Example

okbel:
### Description

This PR updates Feature Flag PostHog Example. Posthog recently added a breaking change for users that are not Pro. Changed strings to booleans. Breaking the example. Fixed in this PR.

### Demo URL

Preview URL

### Type of Change

- [ ] New Example


edge-functions/feature-flag-posthog/lib/posthog-node.ts
```diff
@@ -33,10 +33,14 @@ export async function isFeatureFlagEnabled(
  *          If the feature flag is a multvariate, then the value will be a string
  */
 export async function getFeatureFlagVariant(
-  distinctUserId: string,
+  distinctUserId: string = '',
```
lfades, line 36: Likely unwanted to make this field optional based on the check below about it not being empty

edge-functions/feature-flag-posthog/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+export default async function middleware(req: NextRequest) {
+  const url = req.nextUrl
+
+  if (url.pathname.startsWith('/product')) {
+    // Defaults to product
+    url.pathname = '/product'
+    if (url.pathname != '/product') {
+      // Redirect paths that go directly to the variant
+      return NextResponse.redirect(url)
+    }
```
lfades, line 19: In this case and in the ones similar below, we no longer have to check for the variant or use `startsWith`, the matcher makes sure we're only going to get requests for `/product`

edge-functions/feature-flag-posthog/pages/product/index.tsx
```diff
@@ -4,11 +4,13 @@ export default function Product() {
...
       <Text className="text-lg mb-4">
-        You&apos;re currently looking at the variant of the product page under{' '}
-        <Code>pages/product/c.tsx</Code>
+        This is a fallback. Check middleware.ts&apos;s logic and that
+        PostHog&apos;s SDK correctly set the cookie. You&apos;re currently
+        looking at the variant of the product page under{' '}
+        <Code>pages/product/index.tsx</Code>
       </Text>
       <Link href="/">Go back to /</Link>
     </Page>
```
lfades, line 16: Why was the `c` variant moved to `index`?

lfades (changes requested):
Added some comments, the preview doesn't have a way of accessing the paths in `/products` and that looks unintended, it's either dead code or we're missing links to it.

Also, it should be possible to group everything in `_middleware` into a single check as every flag uses very similar code.

lfades (approved):
I've simplified the middleware.ts implementation to make it more friendly and also the posthog api around it.

The product page now works just like the other pages because its flag is returning a `boolean` instead of a `string` with the variant name, from what the example was showing before it returned a `string` because that one was doing multivariate testing, but that's not what it's returning now and I don't have access to the dashboard to check that.

I've also removed the usage of the API client-side, it should continue to be able to work client-side but its usage in this case was pointless (not a good case for client-side feature flags), we can add it again later if we want to extend the example with that use case, otherwise it's better to skip that CLS for now, plus it gets blocked by adblockers and locally it breaks due to that (that's why I was not seeing in the live deployment, due to me using Brave). cc @okbel

lfades (approved):

PR #340, opened 2022-06-27
follow new middleware used in platforms

GuiBibeau:
### Description

Follows closely the middleware implementation of the main platforms example

https://github.com/vercel/platforms/blob/main/middleware.ts

### Type of Change

- [ ] New Example

### New Example Checklist


lfades (approved):
I'll approve as it looks okay but there's no proper way to test this demo as there's no active demo, so I'm unsure if it works or not

PR #341, opened 2022-06-27
Update optimizely example

goncy:
### Type of Change

- [ ] New Example

### New Example Checklist


lfades (approved):
Beautiful 💯 

I don't think the example works well, I don't know how to test the variants as I always see the same order in the products that it showcases, so it's basically an static page with no dynamic variants. However the current production domain behaves in the same way so it looks unrelated to the changes here.

PR #371, opened 2022-07-29
New Streaming Example

okbel:
### Description
This PR adds a new example using Streams with Vercel's Edge Runtime and Vercel Edge Functions.
The example shows how to modify, transform and work with Streams using external sources. In this case it's using an external gist.

### Type of Change

- [X] New Example

### New Example Checklist


edge-functions/streaming-response/middleware.ts
```diff
@@ -0,0 +1,11 @@
+import { NextResponse, NextRequest } from 'next/server'
+
+export const config = {
+  matcher: '/',
+}
+
+export function middleware(req: NextRequest) {
+  const url = req.nextUrl
+  url.pathname = '/api/edge'
```
goncy, line 9: Can't this be done in `next.config.js` with rewrites or streaming doesn't work doing it that way?

lfades (dismissed):
Requesting changes for now because the example is missing some misc improvements.

edge-functions/streaming-response/middleware.ts
okbel: Sounds good! Both should be technically equal at some point 👍

PR #380, opened 2022-08-24
WASM API Routes examples

ethomson:
### Description

Adding two Edge API Route examples that use WebAssembly.

### Demo URL

🔜 

### Type of Change

- [x] New Example

### New Example Checklist


edge-api-routes/wasm-hello-world/wasm/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "wasm"
+version = "0.1.0"
+authors = ["Guillaume Bibeau-Laviolette <g.bibeaulaviolette@gmail.com>"]
```
lfades, line 4: We shouldn't include personal emails in examples

```suggestion
```

ethomson: Thanks, I squashed this in

edge-api-routes/wasm-hello-world/wasm/pkg/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "wasm",
+  "collaborators": [
+    "Guillaume Bibeau-Laviolette <g.bibeaulaviolette@gmail.com>"
```
okbel, line 4: Please, remove personal emails. Codeowners are already set.

edge-api-routes/wasm-hello-world/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "wasm-hello-world",
+  "devDependencies": {
+    "vercel": "^28.1.3"
```
okbel, line 4: We are at Version 28.1.4 just in case

okbel (changes requested):
Left some comments and I think it might be valuable to rename the examples from `wasm-xor` to `wasm-rust-xor` so when clicking the example users will have a hint that it uses rust.
(As we will add other languages go, python, ruby etc)

Other than that, I believe LGTM!

lfades (approved):

PR #390, opened 2022-09-09
feat(example): add nx example

tknickman:
### Description

Adds a new Next.js example with Nx.

Supports single button deploy for an Nx Next.js app on Vercel.

Migrated from https://github.com/vercel/next.js/pull/40327

### Demo URL

https://solutions-nx-monorepo.vercel.sh

### Type of Change

- [X] New Example

### New Example Checklist

- [X] 📚 The template wasn't used but I carefuly read the [Adding a new example](https://github.com/vercel/examples#adding-a-new-example) steps and implemented them in the example


solutions/nx-monorepo/README.md
```diff
@@ -0,0 +1,109 @@
...
+css: None
+deployUrl: https://vercel.com/new/git/external?repository-url=https://github.com/vercel/examples/tree/main/solutions/nx-monorepo&project-name=nx-monorepo&repository-name=nx-monorepo&output-directory=dist%2Fapps%2Fapp%2F.next&build-command=npx%20nx%20build%20app%20--prod&ignore-command=npx%20nx-ignore%20app%20
+demoUrl: https://solutions-nx-monorepo.vercel.sh
+---
+
+# Nx Monorepo
+
+This is a monorepo example with a single Next.js site ([./app](./app)).
+
+This example was created using [Nx](https://nx.dev).
```
lfades, line 16:
```suggestion
This is a monorepo example using [Nx](https://nx.dev) and a single Next.js site in [./apps/app](./apps/app).
```

solutions/nx-monorepo/apps/app/next.config.js
```diff
@@ -0,0 +1,17 @@
...
+ **/
+const nextConfig = {
+  nx: {
+    // Set this to true if you would like to to use SVGR
+    // See: https://github.com/gregberge/svgr
+    svgr: false,
+  },
+};
+
+module.exports = withNx(nextConfig);
```
lfades, line 17: What does `@nrwl/next/plugins/with-nx` do? I couldn't find any documentation about it, I'm a bit scared this can break with the next major of Next.

solutions/nx-monorepo/babel.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "babelrcRoots": ["*"]
+}
```
lfades, line 3: Is this required in order to have Nx working even if the Next.js app doesn't use babel?

solutions/nx-monorepo/nx.json
```diff
@@ -0,0 +1,48 @@
...
+      "runner": "nx/tasks-runners/default",
+      "options": {
+        "cacheableOperations": ["build", "lint", "test", "e2e"]
+      }
+    },
+    "cloud": {
+      "runner": "@nrwl/nx-cloud",
+      "options": {
+        "cacheableOperations": ["build", "lint", "test", "e2e"],
+        "accessToken": "YOUR ACCESS TOKEN HERE"
```
lfades, line 25: Is this token private? if that's the case then does Nx support this being an environment variable/secret?

solutions/nx-monorepo/workspace.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "./node_modules/nx/schemas/workspace-schema.json",
+  "version": 2,
+  "projects": {
+    "app": "apps/app"
+  }
+}
```
lfades, line 7: Is this file Nx specific or part of one of the package managers?

lfades (changes requested):
Added multiple comments and here are some questions:

- There's no `workspaces` in `package.json` or `pnpm-workspace.yaml`, does the example work with all package managers out of the box?
- Why is `@nrwl/next` not installed as a dev dependency?

Also, can you add the turbo team as the owner of this example in `.github/CODEOWNERS` 🙏

solutions/nx-monorepo/workspace.json
tknickman: This is `Nx` specific - nx has their own notion of workspaces and doesn't rely on the package manager.

solutions/nx-monorepo/nx.json
tknickman: Good call. It _can_ be, often times a read only token will be included here which is safe to publish. But there is an env variable that can be set that takes precedence over this setting - so I'll go that route and include that in a `.env.example`.

This also isn't required, you can just use local caching (the default in this example), but I wanted to include this here because when deploying on Vercel the _version_ of `nx-cloud` that is used matters ([details here](https://github.com/nrwl/nx/issues/10256#issuecomment-1233292721)) so I wanted to have a clear example to point people too.

solutions/nx-monorepo/babel.config.json
tknickman: Good question, let me play with that a little (this came from the Nx project generator)

solutions/nx-monorepo/apps/app/next.config.js
tknickman: Here's the source: https://github.com/nrwl/nx/blob/master/packages/next/plugins/with-nx.ts - it does some webpack shenanigans for environment variable detection to ensure proper cache detection, monorepo, and workspace setup.|

Here's some loose docs as well: https://nx.dev/packages/next

solutions/nx-monorepo/apps/app/.eslintrc.json
```diff
@@ -0,0 +1,31 @@
...
+      "rules": {}
+    },
+    {
+      "files": ["*.js", "*.jsx"],
+      "rules": {}
+    }
+  ],
+  "rules": {
+    "@next/next/no-html-link-for-pages": "off"
+  },
```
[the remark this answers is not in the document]
tknickman: This is what ships with their next boilerplate as well, but let me see if i can simplify some of this

tknickman:
@lfades comments addressed, thanks!

As for your two other questions - Nx has it's own notion of workspaces and doesn't rely on the package manager for this. Also I'm not sure why `@nrwl/next` isn't installed as a dev dependency - there's a possibility this is an error on their app generation side because I can't see a reason why it needs to be there - but I might leave it where they suggest for now until I can figure out the reason.

lfades (approved):
LGTM!

PR #399, opened 2022-09-18
Add Turborepo and remove `next-transpile-modules` usage from @vercel/examples-ui

lfades:
### Description

- The root of our examples is now using Turborepo to handle the packages in the `packages` folder
- GH workflows have been updated to run scripts under `turbo` and remote caching is enabled for faster actions!
- `next-transpile-modules` has been removed from `@vercel/examples-ui`. It's currently being used by most of our examples, we'll start moving away from it and because our examples UI library uses Tailwind, it's fine to rely on the `content` prop of Tailwind to figure out the CSS to use.
- `@vercel/examples-ui` will now release with `changesets`, and it will publish automatically with the `Release` workflow

### Demo URL

The new version hasn't been published so there's no demo yet and the deployment will be failing until then.

### Type of Change

- [ ] New Example


packages/ui/src/index.tsx
```diff
@@ -6,7 +6,6 @@ export { default as Text } from './text'
 export { default as Input } from './input'
 export { default as Button } from './button'
 export { default as LoadingDots } from './loading-dots'
-export { default as A } from './anchor'
```
goncy, line 9: Are this barrels correctly tree-shaken?

packages/ui/src/link.tsx
```diff
@@ -1,13 +1,20 @@
...
 const Link: FC<
   Exclude<LinkProps, 'passHref'> & AnchorHTMLAttributes<HTMLAnchorElement>
 > = ({ href, prefetch, children, className, ...props }) => (
   <NextLink href={href} prefetch={prefetch}>
-    <a className={cn(s.root, className)} {...props}>
+    <a
+      className={cn(
+        'text-link hover:text-link-light transition-colors no-underline',
+        // CSS for <code/>
+        '[&_code]:text-link [&_code]:hover:text-link-light [&_code]:transition-colors',
```
goncy, line 13: Nice

packages/ui/src/tailwind.cjs
```diff
@@ -0,0 +1,97 @@
+const colors = {
```
goncy, line 1: Can't this be a .mjs file? Just by preference

packages/ui/src/globals.css
```diff
@@ -43,7 +35,7 @@ body {
 
 html,
 body {
-  font-family: var(--font-sans);
+  font-family: theme(fontFamily.sans);
```
goncy, line 38: Nice, where is this coming from?

goncy (changes requested):
nits

packages/ui/src/index.tsx
lfades: They should be, yes, the package esports ESModules and Webpack should be able to tree shake it without issues

packages/ui/src/tailwind.cjs
lfades: This file is imported by the `tailwind.config.js` of the app, as commonjs which is what Tailwind allows. I could add another entrypoint with different config to SWC to turn this file from `.mjs` into `.cjs` but that sounds like an unnecessary step.

packages/ui/src/globals.css
lfades: It's the `theme()` directive included by Tailwind: https://tailwindcss.com/docs/functions-and-directives#theme

tknickman (approved):
changesets looks good!

goncy (approved):

PR #406, opened 2022-09-24
Fix ES Modules in new version of vercel/examples package

lfades:
### Description

- For Node.js to properly resolve relative imports they need to have the `.js` extension
- Similarly, if the package imports a js file from a module that doesn't use `exports`, it also has to include the extension, so `next/head` becomes `next/head.js`
- Only one example `ab-testing-simple` has been updated to the latest version of the UI package that's using SWC, I've changed the version in all other packages from `latest` to `0.3.6` which is the last one that used `next-transpile-modules`.
- The `pnpm-lock.yaml` of the UI package was removed because the repo is a monorepo now

This changes continue to be part of the migration away from `next-transpile-modules` and to support the upcoming Next.js release.

### Type of Change

- [ ] New Example

### New Example Checklist


dominiksipowicz (approved):
lgtm

PR #411, opened 2022-09-27
Edited env vars for platforms-related examples to not use VERCEL_ prefix

steven-tey:
### Type of Change

- [ ] New Example
- [ ] Example updates (Bug fixes, new features, etc.)
- [ ] Other (changes to the codebase, but not to examples)

### New Example Checklist

- [ ] 🛫 `npm run new-example` was used to create the example
- [ ] 📚 The template wasn't used but I carefuly read the [Adding a new example](https://github.com/vercel/examples#adding-a-new-example) steps and implemented them in the example
- [ ] 📱 Is it responsive? Are mobile and tablets considered?

PR #413, opened 2022-09-28
Fix  .env.example to reflect real env variable in static-tweets-tailwind example

fordnox:
Fix name of env variable

### Description

App env.example contains invalid ENV name, just renaming env.example

### Demo URL

https://github.com/vercel/examples/blob/main/solutions/static-tweets-tailwind/lib/twitter.js#L20
https://github.com/vercel/examples/blob/main/solutions/static-tweets-tailwind/lib/twitter-media.js#L7


lfades:
Hi there and sorry for the late review 👋 

I'm going to close the PR as the change was merged in https://github.com/vercel/examples/pull/411. Thank you!

PR #420, opened 2022-10-08
Fix typo in user agent edge function example

ps-89:
### Description

I was going through the user agent edge function example and happened to come across a typo, just raising a fix for that.


lfades (approved):
Thank you!

PR #432, opened 2022-10-25
Update monorepo examples to not use next-transpile-modules

lfades:
### Description

- Removed the usage of `next-transpile-modules` from examples with a monorepo
- Moved said examples to `pnpm` from `npm`
- Added other code improvement and dependency updates
- The Vercel project settings of our examples have been updated for the demos to work with this changes

The storybook package in the microfrontends example doesn't work with `pnpm` which is a shame, it's not really important for the example so it can be removed later or we can wait for an update in the Vite builder of Storybook.

### Type of Change

- [ ] New Example


liamlevesqueP8:
so does this mean I can stop transpiling my shared packages in the nextjs app? That would be rad! It looks like I may need a .swcrc file in my packages to make it work? Do you have any resources with more details??

solutions/microfrontends/packages/acme-design-system/package.json
```diff
@@ -4,45 +4,69 @@
...
   },
   "devDependencies": {
+    "@swc/cli": "^0.1.57",
+    "@swc/core": "^1.3.10",
     "@types/react": "latest",
+    "chokidar": "^3.5.3",
     "eslint": "^8.16.0",
-    "eslint-config-acme": "*",
-    "typescript": "^4.5.5"
+    "eslint-config-acme": "workspace:^0.0.1",
```
cond0r, line 69: I prefer here (and other occurrences) `workspace:*`, see turbo [example](https://github.com/vercel/turbo/blob/main/examples/with-pnpm/apps/web/package.json) & [docs](https://pnpm.io/workspaces#publishing-workspace-packages)

cond0r (approved):

lfades:
@liamlevesqueP8 SWC is a way of doing it, you could also use tsup (it uses esbuild) or vite. I choose SWC because it's very fast and easy to setup, although tsup I believe is faster to setup.

For more details the [SWC docs](https://swc.rs/docs/usage/cli) are the way to go.

solutions/microfrontends/packages/acme-pages/package.json
```diff
@@ -28,24 +31,35 @@
...
+        ]
+      }
+    }
+  },
   "peerDependencies": {
-    "next": "^12.1.0",
-    "react": "^17.0.0",
-    "react-dom": "^17.0.0"
+    "next": "*",
+    "react": "^17.0.0 || ^18.0.0-0",
```
goncy, line 46: Do we still need 17?

goncy (approved):

solutions/microfrontends/packages/acme-pages/package.json
lfades: It's okay for the purposes of the example, which is showing how to have shared packages, and these might allow a certain version to be used, even if 17 is not used in the example.

solutions/microfrontends/packages/acme-design-system/package.json
lfades: Done!

PR #437, opened 2022-10-31
Update examples to not include anchor tags in Next.js links

lfades:
### Description

With Next.js 13 having an anchor when using the Link component is not allowed unless the `legacyBehavior` prop is used. This PR updates all links and also the `@vercel/examples-ui` package which also uses Next.js Link.

The UI package will be updated after this PR is merged, once the GH action creates a PR to update the package, after that I'll create another PR to update its version in all examples.

### Type of Change

- [ ] New Example


goncy (approved):

edge-functions/next-news/components/stories.tsx
```diff
@@ -14,7 +14,7 @@ const Stories = ({ stories, page = 1, offset = null }) => (
       </div>
     ))}
     <footer className="footer">
-      <Link href={`/news/${page + 1}`}>
+      <Link href={`/news/${page + 1}`} legacyBehavior>
         <a>More</a>
```
jaredpalmer, line 18: Why legacy here and not codemod?

PR #442, opened 2022-11-02
use `legacyBehavior` in link exported by the examples UI package.

lfades:
### Description

There are some styling issues with the link component in the package when used in multiple examples, with this change it's going back to how it was being used before.

### Type of Change

- [ ] New Example

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -9,7 +9,7 @@
     "clean": "rm -rf ./dist",
     "types": "tsc --emitDeclarationOnly"
   },
-  "sideEffects": false,
+  "sideEffects": true,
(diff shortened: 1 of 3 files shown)


packages/ui/package.json
```diff
@@ -9,7 +9,7 @@
     "clean": "rm -rf ./dist",
     "types": "tsc --emitDeclarationOnly"
   },
-  "sideEffects": false,
+  "sideEffects": true,
```
lfades, line 12: The package has side effects because it exports a css file.

cond0r (approved):

goncy (approved):

PR #448, opened 2022-11-08
Update loading-web-fonts example to use the new and improved Font Optimization of Next.js

lfades:
Demo URL: [solutions-loading-web-fonts-bkpkni1fq.vercel.sh](http://solutions-loading-web-fonts-bkpkni1fq.vercel.sh/)

### Type of Change

- [ ] New Example


cond0r (approved):

solutions/loading-web-fonts/package.json
```diff
@@ -9,8 +9,9 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@next/font": "^13.0.2",
     "@vercel/examples-ui": "^1.0.4",
-    "next": "latest",
+    "next": "canary",
```
cramforce, line 14: Why?

solutions/loading-web-fonts/pages/index.tsx
```diff
@@ -61,23 +77,33 @@ function Home() {
...
+          <Link
+            href="https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/size-adjust"
+            target="_blank"
+          >
+            <Code>size-adjust</Code>
+          </Link>{' '}
+          set to <b>107%</b>:
         </Text>
-        <FontMixer fonts={['Inter', 'Inter-fallback']}>
+        <FontMixer fonts={['var(--inter-font)', 'Inter-fallback']}>
```
cramforce, line 106: Not sure whether the fallback works like that, or is it hard coded in the CSS?

solutions/loading-web-fonts/pages/_app.tsx
```diff
@@ -15,7 +37,9 @@ function App({ Component, pageProps }: AppProps) {
       path="solutions/loading-web-fonts"
       description="How to correctly load web fonts"
     >
-      <Component {...pageProps} />
+      <div className={inter.variable}>
```
cramforce, line 40: I don't love advertising the variable, but it definitely makes sense for this example!

solutions/loading-web-fonts/pages/index.tsx
lfades: There's a hard coded fallback font in a CSS file, the `FontMixer` component shows a comparison of two fonts, and because I don't know the name of the generated fallback font we use a predefined one here, but that's only used internally for the purposes of the example.

solutions/loading-web-fonts/pages/_app.tsx
lfades: Thinking about it I think this could be improved!

solutions/loading-web-fonts/package.json
lfades: `canary` is never supposed to have breaking changes and so we set examples to it to have them with the latest changes when cloned, this is all optional and we have had cases where using `canary` would be needed to fix a bug, and also where a breaking change was introduced accidentally by Next.js, and this helps catching that.

solutions/loading-web-fonts/pages/_app.tsx
lfades: Improved with the last commit, the example is now using `className`.

PR #455, opened 2022-11-15
Fix types in `useRouter` usage for multiple examples

lfades:
### Description

There was a canary version that required to do `useRouter(true)` for some cases, but that's no longer required. 

### Type of Change

- [ ] New Example


okbel (approved):
SGTM

PR #460, opened 2022-11-22
fix(monorepo): turbo schema and package manager

tknickman:
### Description

Fix schema URL and add required packageManager field.

### Type of Change

- [ ] New Example

### New Example Checklist


tknickman:
@lfades updated, I also included the lockfile for the monorepo example because turbo needs this to use `turbo-ignore`

PR #463, opened 2022-11-23
Update middleware.ts

theindianappguy:
without this / all the subdomain were showing the same data

### Description

So i am adding subdomain feature to my app while using this i faced issue all the subdomain were showing the same data on investigating i realised / is missing which fixed the issue for me.

### Demo URL

https://subdomain-2.early.sbs

### Type of Change

Other (changes to the codebase, but not to examples)

### New Example Checklist

NA


lfades:
`url.pathname` always includes a forward slash so adding this is not needed and might break something, I tested it out locally to confirm.

PR #467, opened 2022-12-01
Move testing solution to a monorepo and pnpm

lfades:
### Description

This moves `solutions/testing` to a monorepo with Turborepo, and now the test and build scripts will have cache enabled. The example is also now setup for `pnpm` only and the `pnpm-lock` file is committed to git.

### Type of Change

- [ ] New Example


dominiksipowicz (approved):
lgtm

goncy (approved):

PR #469, opened 2022-12-02
Add Python examples.

leerob:
This adds three examples for a hello world Python function, Django, and Flask using Vercel Serverless Functions.

https://flask-python-template.vercel.app/
https://django-template.vercel.app/
https://python-hello-world.vercel.app/


leerob:
I've deployed these all under Vercel Examples and confirmed they're working:

- https://flask-python-template.vercel.app/
- https://django-template.vercel.app/
- https://python-hello-world.vercel.app/

python/hello-world/README.md
```diff
@@ -0,0 +1,24 @@
...
+This example shows how to use Python on Vercel with Serverless Functions using the [Python Runtime](https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/python).
+
+## Demo
+
+https://python-hello-world.vercel.app/
+
+## Running Locally
+
+```bash
+npm i -g vercel
```
dominiksipowicz, line 14:
```suggestion
pip install vercel
```

I think for python it would be cool to do use `pip`

https://pypi.org/project/vercel/

not sure how updated those are and what's out strategy here. So I leave it as an optional suggestion.

python/flask/vercel.json
```diff
@@ -0,0 +1,14 @@
...
+      "use": "@vercel/python"
+    }
+  ],
+  "routes": [
+    {
+      "src": "/(.*)",
+      "dest": "api/index.py"
+    }
+  ]
+}
```
dominiksipowicz, line 14: Similar to other examples, do we want redirect from index to API route?

dominiksipowicz (approved):
Added some minor comments. lgtm.

python/flask/vercel.json
leerob: So to get Flask / Django to work, we had to customize the builder through `vercel.json` - this handles setting up the function to render for the defined routes, in this case the index route.

PR #473, opened 2022-12-13
add missing folders to pnpm-workspace

dferber90:
### Description

When I ran `pnpm i` on the root it didn't install the dependencies of examples nested under `edge-functions`.

So when I then ran `pnpm dev` under `edge-functions/ab-testing-simple` it failed due to missing dependencies.

Running `pnpm i` under `edge-functions/ab-testing-simple` didn't install the dependencies either as pnpm seems to add them at the root instead.

This PR adds all folders to `pnpm-workspace.yaml` so that their dependencies get installed as well when running `pnpm i`.

### Type of Change


pnpm-workspace.yaml
```diff
@@ -1,2 +1,7 @@
 packages:
   - 'packages/*'
+  - 'app-directory/*'
+  - 'edge-api-routes/*'
+  - 'edge-functions/*'
+  - 'plop-templates/*'
+  - 'solutions/*'
```
dominiksipowicz, line 7: Those changes probably aren't what we want. All of the examples are not mono repo packages, but examples on their own including **mono repo examples**. Deps shouldn't be installed globally for every example. `/packages/*` folder contains the actual packages we want to control. 

When people want to try an example, they will only install the dependencies for that example. In addition, we have Python examples that will not be served by `pnpm`

dferber90: > they will only install the dependencies for that example

can you explain how to do this? running `pnpm i` in any folder tries to install them at the root

dominiksipowicz (approved):

pnpm-workspace.yaml
dominiksipowicz: I was running `npm i` inside the examples as it is in the readme but yes, it looks like it's broken for `pnpm` when run inside the examples.

dferber90:
Closing this as we should likely add `pnpm-workspace.yaml` files to each example as this repo actually is not a single workspace but rather a collection of workspaces

PR #496, opened 2022-12-21
Create share-state example

goncy:
### Description

image

Shows how to share state between layouts and pages

### Demo URL

https://goncy-app-dir-share-state.vercel.app/

### Type of Change

- [x] New Example

### New Example Checklist

- [x] 🛫 `npm run new-example` was used to create the example


app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,178 @@
...
+}`}
+        </Snippet>
+      </section>
+
+      <section className="flex flex-col gap-3">
+        <Text variant="h2">Fetching data from the layout</Text>
+        <Text>
+          If you need your layout to fetch information from the server, you can
+          create a group and make the layout inside the group a client component
+          instead of the root layout
```
lfades, line 155: This says "and make the layout inside the group a client component", but if the group is `(app)` then it doesn't have a layout.

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const state = useState(0);
+
+  return <CounterContext.Provider value={state}>{children}</CounterContext.Provider>;
+}
+
+function useCounter() {
+  return useContext(CounterContext);
+}
+
+export { CounterProvider, useCounter };`}
```
lfades, line 61: This use of useState is fine but it's not optimized for the use case that is being displayed in the demo, where the Layout **reads** and the pages **write**, because using the `useCounter` hook automatically subscribes you to both.

Here's a more optimized variant of this for a functionality that saves the current hash of a page and that can be modified for the counter:

```ts
import {
  useContext,
  useState,
  createContext,
  type Dispatch,
  type SetStateAction,
} from 'react';

type SetActiveHash = Dispatch<SetStateAction<string>>;

const ActiveHash = createContext('');
const SetActiveHash = createContext<SetActiveHash>(() => '');

export function ActiveHashProvider({
  children,
}: {
  children: React.ReactNode;
}): JSX.Element {
  const [activeHash, setActiveHash] = useState('');

  return (
    <ActiveHash.Provider value={activeHash}>
      <SetActiveHash.Provider value={setActiveHash}>
        {children}
      </SetActiveHash.Provider>
    </ActiveHash.Provider>
  );
}

export const useActiveHash = (): string => useContext(ActiveHash);

export const useSetActiveHash = (): SetActiveHash => useContext(SetActiveHash);
```

app-directory/share-state/.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "typescript.tsdk": "node_modules\\typescript\\lib",
+  "typescript.enablePromptUseWorkspaceTsdk": true
+}
```
lfades, line 4: Do we need this file?

lfades (changes requested):
Added some comments

app-directory/share-state/.vscode/settings.json
goncy: Is added automatically when you agree to use the Next.js plugin, we can add the `.vscode` folder to gitignore if you want

app-directory/share-state/app/page.tsx
goncy: I agree that its more performante, my concern is to complicate the example or make people question why there are two contexts :thonk:

app-directory/share-state/app/page.tsx
goncy: I forgot to add it to the file table, thanks!

app-directory/share-state/app/page.tsx
lfades: I think that can be solved with in-code comments, but having something like this is better for globally shared state like in this example.

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+|__/nested
+|____/page.js (server)
+`}</pre>
+        <Text>
+          Lets create a simple context that stores a <Code>counter</Code> value
+          and exports a <Code>useCounter</Code> hook to interact with that
+          value.
+        </Text>
+        <Snippet>
+          {`import { createContext, useContext, useState } from "react";
```
dominiksipowicz, line 47:
```suggestion
          {`// app/context.js
import { createContext, useContext, useState } from "react";
```

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+const CounterContext = createContext({});
```
dominiksipowicz, line 49: Would it be possible to add a comment here as to why Context API should be used instead of just using `useState` hook?

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export { CounterProvider, useCounter };`}
+        </Snippet>
+        <Text>
+          Now, lets wrap our root layout with this context provider. This will
+          make the <Code>useCounter</Code> hook available to all the pages
+          inside this layout.
+        </Text>
+        <Snippet>
+          {`"use client";
```
dominiksipowicz, line 69:
```suggestion
          {`// app/layout.js
"use client";
```

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+    </CounterProvider>
+  );
+}`}
+        </Snippet>
+        <Text>
+          Now, lets create a counter component that will allow us to increment
+          and decrement the counter value.
+        </Text>
+        <Snippet>
+          {`"use client";
```
dominiksipowicz, line 104:
```suggestion
          {`// app/counter.js
"use client";
```

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+    </div>
+  );
+}`}
+        </Snippet>
+        <Text>
+          And finally we will create an index and a nested page with the same
+          content that will render the counter component.
+        </Text>
+        <Snippet>
+          {`import Counter from "./counter";
```
dominiksipowicz, line 125:
```suggestion
          {`// app/page.js
import Counter from "./counter"; 
```

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+  return (
+    <div>
+      <h1>Index page</h1>
+      <Counter />
+    </div>
+  );
+}`}
+        </Snippet>
+        <Snippet>
+          {`import Counter from "../counter";
```
dominiksipowicz, line 137:
```suggestion
          {`// app/nested/page.js
import Counter from "../counter"; 
```

app-directory/share-state/app/page.tsx
```diff
@@ -0,0 +1,179 @@
...
+|______/page.js (server)
+`}</pre>
+      </section>
+
+      <section className="flex flex-col gap-3">
+        <Text variant="h2">Demo</Text>
+        <Text>
+          Now you will be able to fetch information from the server while
+          sharing state between layouts and routes, you can see a working demo
+          in <Link href="/demo">here</Link>.
```
dominiksipowicz, line 174:
```suggestion
          <Link href="/demo">here</Link>.
```

lfades (approved):

PR #505, opened 2023-01-03
Create an HTML only example.

leerob:
I'm not sure if `framework: none` will work for Contentful, hmmm.

### DIFF
--- a/solutions/html/README.md
+++ b/solutions/html/README.md
@@ -1,0 +1,13 @@
+---
+name: HTML Starter
+slug: html-starter-with-analytics
+description: HTML5 template with analytics and advanced routing configuration.
+framework: None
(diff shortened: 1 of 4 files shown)


solutions/html/README.md
```diff
@@ -0,0 +1,13 @@
+---
+name: HTML Starter
+slug: html-starter-with-analytics
+description: HTML5 template with analytics and advanced routing configuration.
+framework: None
```
lfades, line 5: Framework is optional but I'm not sure if `None` would work.
```suggestion
```

lfades (changes requested):
The `relatedTemplates` field is required in `Contentful`, we could add a checker for this in the future.

Also, could we name the example with something more descriptive, like `html-starter`?

lfades (approved):

PR #511, opened 2023-01-06
Fix mint nft build

goncy:
Fix build for mint-nft project


okbel:
Works perfectly. Thank you!

okbel (approved):

PR #513, opened 2023-01-09
Major Update Mint NFT Solutions from Moralis v1 to Moralis v2

YosephKS:
### Description

This is an update of Mint NFT solutions example from using Moralis v1 to Moralis v2. Update includes:
- Removal of `moralis-v1`, `react-moralis`, and Moralis Rarible Plugins (all deprecated and not recommended)
- Using Moralis v2 Auth API and Web3 API through `@moralisweb3/next` hooks
- Addition of Hardhat folder for ERC721 NFT contract deployment (Optional)
- Redirection after NFT mint completion will be to OpenSea instead of Rarible
- The requirement of Moralis server has been removed and therefore all backend are now using NextJS backend

Other than that, no changes should be made from the interface level, everything implemented should have the same flow as previous iterations.

### Type of Change

- [x] New Example


solutions/mint-nft/components/Mint.tsx
```diff
@@ -107,16 +128,16 @@ export const Mint: React.VFC = () => {
...
-              This process might take up to 1 minute to complete
+              This process might take up to 1 minute to complete. Please refresh
+              the OpenSea page if you didn&apos;t see your NFT.
             </span>
           </Text>
           <section className="relative w-full pb-[20%] h-48 pb-6 mt-12">
             <Image
               className="rounded-xl"
-              src={String(asset?._url)}
+              src={asset ?? ""}
```
okbel, line 138: Empty src will error out at runtime (linter should show it though) Will it ever be empty? If so, do something like ⬇️  this higher up

```
{ asset ? <Image /> : null }
```

solutions/mint-nft/hardhat/artifacts/@openzeppelin/contracts/token/ERC721/ERC721.sol/ERC721.dbg.json
```diff
@@ -0,0 +1,4 @@
+{
```
okbel, line 1: Are all these ( NFT contract deployment ) necessary and the entire artifacts folder? - We would like to keep the example as minimal as possible for easy adoption

solutions/mint-nft/utils/constant.ts
```diff
@@ -0,0 +1,3 @@
+const NFTAddress = "0x343280B77B62F0DC48243dDFE2D4bC7e4B1f1172";
```
okbel, line 1: Is this a needed placeholder?

okbel (commented):
Tiny changes and some questions. Thank you for the PR!!

solutions/mint-nft/components/Mint.tsx
YosephKS: Well it shouldn't be, but I do have the typing for `asset` to be `string | null` so that might be why it is causing the typing error, let me try your solution

solutions/mint-nft/hardhat/artifacts/@openzeppelin/contracts/token/ERC721/ERC721.sol/ERC721.dbg.json
YosephKS: So different to prior implementation that use lazy minting, this requires on-chain minting and will need those artifacts folder, while they can be generated on the fly, they are generally pushed (not put in .gitignore) to the repo so the ABI can be used in production to communicate with the deployed smart contract on the client side

As far as implementation this should be pretty simple, artifacts might look complicated but most of the time it is unnecessary for user to really know what's inside (kinda like `node_modules`), they just need to know that the ABI is there and they can search for the ABI when they need them

solutions/mint-nft/utils/constant.ts
YosephKS: Yes we need to, in order to mint the NFT, which means writing to the smart contract, we'll need to specify which contract address to write to

PR #515, opened 2023-01-09
fix port unavailable issue when running next start

devsrv:
### Description

running multiple next apps with `pnpm start` enforces all the apps to run on port 3000. causing port not available error for all the apps other than the main app

image

### Solution
running the next apps on unique port.
added `-p 3001` to the start script


lfades (approved):
Thank you!

PR #533, opened 2023-01-26
Github Actions with Vercel CI/CD Example

ah100101:
### Description

This is a new example showing how a user might leverage Vercel CI/CD and add a custom GitHub action for end to end testing (and any additional checks) and also perform a release. The example came up in the context of a customer requesting a best practice for Trump Based Development. The action `.yaml` contains everything needed and the README details the CI/CD process.

Example Readme has full details

### Demo URL

There is a video included at the bottom of the README showing how this example can work

### Type of Change

- [x] New Example

### New Example Checklist

This example only contains an `index.html` file since this is a CI/CD example.

- [x] 🛫 `npm run new-example` was used to create the example
- [N/A] 📱 Is it responsive? Are mobile and tablets considered?

### DIFF
--- a/ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
+++ b/ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
@@ -1,0 +1,17 @@
+name: E2E-Release-Production
+on: [deployment_status]
+jobs:
+  e2e:
+    if: github.event.deployment.environment == 'Production' && github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Placeholder for running e2e tests via Github Actions (Cypress, Playwright, etc)
+        run: sleep 30
+  release:
+    needs: e2e
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Vercel CLI
+        run: npm install --global vercel@latest
+      - name: Alias Production URL
+        run: vercel alias --token=${{ secrets.VERCEL_TOKEN }} set $VERCEL_URL https://github-actions-with-vercel-cicd-production.vercel.app/
--- a/ci-cd/github-actions-with-vercel-cicd/.gitignore
+++ b/ci-cd/github-actions-with-vercel-cicd/.gitignore
@@ -1,0 +1,41 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# Dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# Testing
+/coverage
+
+# Next.js
+/.next/
+/out/
+
+# Production
+build
+dist
+
+# Misc
+.DS_Store
+*.pem
+
+# Debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Local ENV files
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+# Vercel
+.vercel
+
+# Turborepo
+.turbo
+
+# typescript
+*.tsbuildinfo
--- a/ci-cd/github-actions-with-vercel-cicd/README.md
+++ b/ci-cd/github-actions-with-vercel-cicd/README.md
@@ -1,0 +1,66 @@
+---
+name: GitHub Actions with Vercel CI/CD
+slug: github-actions-with-vercel-cicd
+description: Follow a Trunk Based Development workflow with Vercel CI/CD and GitHub Actions for Production testing and release
+---
+
+# Maintain Maximum Flexibility and Platform Features with Vercel CI/CD and GitHub Actions
+
+Use Github Actions to consume the output of Vercel's CI/CD process to run e2e testing, additional checks, and release while also maintaining access to Vercel's built-in CI/CD features like Promote to Production, Checks, and Instant Rollbacks.
+
+This approach is useful for developers and teams that intend on leveraging all of Vercel's features but wish to run additional logic prior to production releases.
+
+# Configuring GitHub Actions for Vercel
+
+In this example there is an action defined at `.github/workflows/e2e-release-production.yaml`:
+
+```yaml
+name: E2E-Release-Production
+on: [deployment_status]
+jobs:
+  e2e:
+    if: github.event.deployment.environment == 'Production' && github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Placeholder for running e2e tests via GitHub Actions (Cypress, Playwright, etc)
+        run: sleep 30
+  release:
+    needs: e2e
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Vercel CLI
+        run: npm install --global vercel@latest
+      - name: Alias Production URL
+        run: vercel alias --token=${{ secrets.VERCEL_TOKEN }} set $VERCEL_URL https://github-actions-with-vercel-cicd-production.vercel.app
+```
+
+This action:
+
+1. Runs on the `deployment_status` change. This lets Vercel run its default CI/CD process and features.
+2. Will only run when the environment is `Production` and the `github.event.deployment_status.state` is successful.
+3. Runs an `e2e` job with any testing framework the developer chooses (we are using a placeholder command in this example).
+4. Runs a `release` job upon a successful completion of the `e2e` job.
+5. The `release` job uses the Vercel CLI's `alias` command to alias the production URL (`https://github-actions-with-vercel-cicd-production.vercel.app/`) to the deployment URL (stored in the `$VERCEL_URL` environment variable). Note that on a production project the production URL should be set to a Custom Domain in Vercel's [Domains Dashboard](https://vercel.com/dashboard/domains). This example is using a `.vercel.app` domain for example purposes.
+
+Set the required settings in Vercel and GitHub:
+
+1. If not added already, [add a custom domain on Vercel](https://vercel.com/docs/concepts/projects/domains/add-a-domain)
+2. Retrieve a [Vercel Access Token](https://vercel.com/support/articles/how-do-i-use-a-vercel-api-access-token)
+3. In GitHub, add `VERCEL_TOKEN`, `VERCEL_ORG_ID`, and `VERCEL_PROJECT_ID` as [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets).
+4. Replace the placeholder (or leave for demo purposes) `sleep 30` in the `e2e` workflow job with the test script of your choosing.
+
+# Deploying a Change to Production
+
+Once the above settings are configured, test this workflow by making a change:
+
+1. Create a new pull request in the GitHub repository.
+2. Vercel will create a new build and subsequent Preview Deployment for the Pull Request (skipping the custom action workflow since it is not a Production release).
+3. Upon the successful deployment the changes will be visible on a preview URL for review.
+4. Approve and merge the open Pull Request into `main`.
+5. Vercel will create a new production build and production deployment for the merge into `main`. Note that upon the Vercel deployment completing the changes will be visible on `https://github-actions-with-vercel-cicd.vercel.app/` (the default URL assigned to the project) which is **not** the custom domain configured (which will be done in a following step).
+6. The GitHub Action will run, starting with any configured e2e tests on the non-production or deployment domain (e.g. `https://github-actions-with-vercel-cicd.vercel.app/`).
+7. Upon successful completion of the e2e testing job the release job will run and alias the custom domain (e.g. `https://github-actions-with-vercel-cicd-production.vercel.app/`) to the deployment domain and the change will now be live on the production URL.
+
+## Continue Using Vercel Build and Deployment Features
+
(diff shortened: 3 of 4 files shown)


ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
```diff
@@ -0,0 +1,17 @@
+name: E2E-Release-Production
```
dominiksipowicz, line 1: The purpose of this comment is simply to provide information for discussion; no action is required.

Would it be possible to create a second file for feature branches only so that it will be clear that we want separate flow for preview and production? 

We still want to run E2E using QA data on PR before merging to production and then we will run E2E using staging or pre-prod/prod data

So the preview yaml file would have something like

```

on:
  push:
    branches-ignore:
    - master
    
```

dominiksipowicz:
I am happy to be merged as it is in its current state. The comment has been added for visibility and discussion.

ci-cd/github-actions-with-vercel-cicd/.gitignore
```diff
@@ -0,0 +1,41 @@
...
+.env.production.local
+
+# Vercel
+.vercel
+
+# Turborepo
+.turbo
+
+# typescript
+*.tsbuildinfo
```
lfades, line 41: A `.gitignore` is not needed.

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - name: Placeholder for running e2e tests via Github Actions (Cypress, Playwright, etc)
+        run: sleep 30
+  release:
+    needs: e2e
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Vercel CLI
+        run: npm install --global vercel@latest
+      - name: Alias Production URL
+        run: vercel alias --token=${{ secrets.VERCEL_TOKEN }} set $VERCEL_URL https://github-actions-with-vercel-cicd-production.vercel.app/
```
lfades, line 17: Doesn't the action need to define its env with the `env` config first?

ci-cd/github-actions-with-vercel-cicd/README.md
```diff
@@ -0,0 +1,72 @@
...
+3. Runs an `e2e` job with any testing framework the developer chooses (we are using a placeholder command in this example).
+4. Runs a `release` job upon a successful completion of the `e2e` job.
+5. The `release` job uses the Vercel CLI's `alias` command to alias the production URL (`https://github-actions-with-vercel-cicd-production.vercel.app/`) to the deployment URL (stored in the `$VERCEL_URL` environment variable). Note that on a production project the production URL should be set to a Custom Domain in Vercel's [Domains Dashboard](https://vercel.com/dashboard/domains). This example is using a `.vercel.app` domain for example purposes.
+
+Set the required settings in Vercel and GitHub:
+
+1. If not added already, [add a custom domain on Vercel](https://vercel.com/docs/concepts/projects/domains/add-a-domain)
+2. Retrieve a [Vercel Access Token](https://vercel.com/support/articles/how-do-i-use-a-vercel-api-access-token)
+3. In GitHub, add `VERCEL_TOKEN`, `VERCEL_ORG_ID`, and `VERCEL_PROJECT_ID` as [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets).
```
lfades, line 49: only the `VERCEL_TOKEN` is used by the action, are the other 2 also used?

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
lfades: Please explain it better, by QA do you mean a normal push that creates a preview deployment? and why do we need staging and how would you do staging?

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
```diff
@@ -0,0 +1,17 @@
+name: E2E-Release-Production
+on: [deployment_status]
+jobs:
+  e2e:
+    if: github.event.deployment.environment == 'Production' && github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
```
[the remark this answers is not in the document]
lfades: Action required, this action targets all branches instead of pushes to `main`.

lfades (changes requested):
Added some comments but overall I don't see why this new example can't just be a PR to https://github.com/vercel/examples/tree/main/solutions/testing

dominiksipowicz:
> Added some comments but overall I don't see why this new example can't just be a PR to https://github.com/vercel/examples/tree/main/solutions/testing
> 

Good question. This example should be named "Trunk Based Development CI/CD workflow example". Testing is not the focus of this example. To simulate testing, it only uses 30 seconds of mock delay.

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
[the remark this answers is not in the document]
ah100101: Correct. The domain should be set in the Custom Domains only. If it's set in the project's domains then it will be assigned a branch it's subsequently deployed via Vercel's CI/CD and break this flow.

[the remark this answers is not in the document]
dominiksipowicz: ok yeah that what I was about. Can we add this to README please

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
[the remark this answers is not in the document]
ah100101: An `if` condition that checks the `github.ref` for the branch is not possible when running on `deployment_status`.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#deployment_status

In fact, there are no properties on `github` that contain the branch name. I am investigating alternatives for running this on a specific branch.

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
ah100101: A second yaml file can be created that runs e2e testing on pre-production environments should a team need to. Below is an example that would run e2e tests in pre-production and does not use `vercel alias`. A merge for a PR will be blocked if an e2e tests(s) fails with this action.

```yaml
name: E2E-Pre-Production
on: [deployment_status]
jobs:
  e2e:
    if: github.event.deployment.environment != 'Production' && github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
    runs-on: ubuntu-latest
    steps:
      - name: Placeholder for running e2e tests via Github Actions (Cypress, Playwright, etc)
        run: [e2e command here]
```

This new example is focused on providing Vercel CI/D with e2e _and_ releasing to production (via `alias`), which is why I've only included an action for production. I can add another for a separate environment if that fills in any gaps we think should be addressed here. 

Lastly, I'll note that it is not possible to check whether a certain branch is deployed on `deployment_status` because the `deployment_status` event is triggered when a deployment status is updated, and not when the deployment is created. The `ref` field typically refers to the branch or tag that is being deployed, and since deployment statuses are not associated with a specific branch or tag, the `ref` field is not included in the `deployment_status` event payload. 
However, `github.event.deployment.environment` is available and will be set to Production, Preview or Development, which is leveraged in this example.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#deployment-status
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#deployment

ah100101:
> Added some comments but overall I don't see why this new example can't just be a PR to https://github.com/vercel/examples/tree/main/solutions/testing
> 
> Update: can the use case here be explained a bit more? production is ready once a PR to the main branch is merged and the deployment for it gets created, the action here is running tests over that branch and then adding another alias on the same deployment if the tests worked, how is that an improvement over running the tests over the preview branch before merged and not allowing a merge unless E2E works? The production deployment won't have any additional changes that can justify running E2E again before releasing IMO, if that were needed then that would be the use case of "staging".

@lfades 
The justification for this use case is associated with having a Trunk Based Development workflow. For scenarios where e2e tests run against a Preview deploy prior to merging into main, Preview may be using non-production data (e.g. content, env variables, pre-production services), and some teams may want to test closer to the production end user experience. Testing e2e on a staging environment is one way to accomplish this, but some teams following a Trunk Based workflow may require testing that is on the production environment. This example shows another version, similar to a blue-green deployment, where e2e tests happen directly on a Production deployment and when all tests pass the public domain is aliased to that deployment URL that was previously inactive / inaccessible to end users.

If you think that additional context is helpful here I will update the readme.

lfades:
@ah100101 Rename the example and update the readme. Having the action aliasing to production is not ideal, if checks fail the reporting is less useful because there's no open PR, and not having a PR open also works badly with third party apps that do checks in GH PRs, so having a middle branch that's also production can work better. That branch can make automatic PRs and update them when something new is merged into the branch, so that you can properly follow up fixes. But I'm okay with taking what's in here for now 👍

ci-cd/github-actions-with-vercel-cicd/.gitignore
ah100101: This has been removed.

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
ah100101: I've updated the the alias line to use GitHub Action variables and have updated the readme to account for this change. An `env` does not need to be defined for $VERCEL_URL since that is a Vercel System Environment Variable. I've linked the readme to the docs.

ci-cd/github-actions-with-vercel-cicd/README.md
ah100101: I have removed VERCEL_ORG_ID and VERCEL_PROJECT_ID

ci-cd/github-actions-with-vercel-cicd/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Continue Using Vercel Build and Deployment Features
+
+Since the Vercel CI/CD process is still utilized in this example teams can continue to leverage [Instant Rollback](https://vercel.com/docs/concepts/deployments/instant-rollback), [Promote to Production](https://vercel.com/docs/concepts/deployments/instant-rollback#instant-rollback-vs.-promote-to-production), and [Checks](https://vercel.com/docs/integrations/checks-overview) as designed.
+
+# Demo
+
+
+https://user-images.githubusercontent.com/6125972/214726723-27039433-efa8-4f0e-895a-71fc5cc0319e.mov
+
```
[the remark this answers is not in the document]
ah100101: Good catch that this is confusing. I've uploaded a new video with a more clear URL naming convention.

ci-cd/github-actions-with-vercel-cicd/.github/workflows/e2e-release-production.yaml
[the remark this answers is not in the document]
ah100101: I've updated the readme with an additional note about not assigning in the Project's Domain and only the Custom Domains with a link.

lfades (commented):
I think a good and not very long name can be `github-trunk-based-development`

ah100101:
> @ah100101 Rename the example and update the readme. Having the action aliasing to production is not ideal, if checks fail the reporting is less useful because there's no open PR, and not having a PR open also works badly with third party apps that do checks in GH PRs, so having a middle branch that's also production can work better. That branch can make automatic PRs and update them when something new is merged into the branch, so that you can properly follow up fixes. But I'm okay with taking what's in here for now 👍

@lfades Circling back to this PR, the example has been renamed and readme updated. Are there more changes you would like to see done here? Thanks!

ci-cd/github-trunk-based-development/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+Set the required settings in Vercel and GitHub:
+
+1. [Add a custom domain on Vercel](https://vercel.com/docs/concepts/projects/domains/add-a-domain). Do not add the domain under the Project Settings Domains. Doing so will require assigning a branch and deployments will assign that URL prior to the `alias` step and that is not desired for this workflow.
+2. Retrieve a [Vercel Access Token](https://vercel.com/support/articles/how-do-i-use-a-vercel-api-access-token)
+3. In GitHub, add `VERCEL_TOKEN` as a [secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets).
+4. In GitHub, add `PRODUCTION_URL` as a [variable](https://docs.github.com/en/actions/learn-github-actions/variables).
+5. Replace the placeholder (or leave for demo purposes) `sleep 30` in the `e2e` workflow job with the test script of your choosing.
```
lfades, line 51: This is bad UX. Is there a way to improve this?

lfades (changes requested):
The usage of the term "trunk based development" should be removed in my opinion as it's something Vercel does by default, and not something that can only be done with this example. I can be wrong here but after a second look at what trunk based development means that term might not be good for us.

leerob:
Isn't the term "trunk based development" the whole point of the example, though?

dominiksipowicz:
To make sure we refer to one single branch, it would be very helpful for users if the name of this example included **"trunk-based".** It is possible to use other names, such as: 

- **Advanced Trunk Based Development** (true since we run extra checks on new deployments before promoting them to production)
- **Gated trunk-based deployments** (same as above)
- **Blue/green Trunk Based deployments** (true since we have two deployments running at the same time)

WDYT?

lfades:
advanced trunk based development sounds good to me.

PR #540, opened 2023-01-31
Updates README error message when syncing templates to Contentful.

manovotny:
### Description

Needlessly checking for a `package.json` file when trying to read `README.md` content. 

The `README.md` frontmatter should take care of marketplace availability (ie. `marketplace: false`).

### Type of Change

### DIFF
--- a/scripts/lib/contentful/update-template.ts
+++ b/scripts/lib/contentful/update-template.ts
@@ -16,7 +16,7 @@ export default async function updateTemplate({
   const readme = await getReadme(examplePath)
 
   if (!readme) {
-    throw new Error('No readme.md found in example directory')
+    throw new Error(`No README.md found in example directory ${examplePath}`)
   }
 
   const { body: readmeBody, lang, template } = await getTemplate(readme)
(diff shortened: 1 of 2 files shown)


steven-tey (approved):
Looks good to me!!

steven-tey (commented):
we're throwing an error here: https://github.com/vercel/examples/blob/updates-readme-error-message/scripts/lib/get-readme.ts#L7-L9

Doesn't that mean that we should be catching it in the top level via a `try...catch`?

scripts/lib/contentful/update-template.ts
```diff
@@ -16,7 +16,7 @@ export default async function updateTemplate({
   const readme = await getReadme(examplePath)
```
steven-tey, line 16: Here's the top level I was referring to in my previous comment

manovotny: I can add that, however, we weren't catching it before this PR either.

manovotny: Actually, do we want to catch it? We want the script to fail, so I don't think we do.

steven-tey: ahh okay fair enough! In that case let's merge it! :)

PR #544, opened 2023-02-01
Updated Static Tweets Example and converted it to Typescript

steven-tey:
Related: this example was throwing errors and blocking #541

### Type of Change

- [ ] New Example

### New Example Checklist


solutions/static-tweets-tailwind/lib/remark-plugins.tsx
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      resolve()
+    })
+}
+
+async function getTweet(node: Literal<string>) {
+  const regex = /\/status\/(\d+)/gm
+
+  const matches = regex.exec(node.value)
```
styfle, line 58: Can you use `new URL(node.value)` to parse it instead of regex?

steven-tey: I believe it's a string

styfle (approved):

PR #546, opened 2023-02-02
fix: upgrade `next-auth` to fix TS issue

balazsorban44:
See https://github.com/nextauthjs/next-auth/pull/6595 for more context


steven-tey (approved):
LGTM THANK YOU!

PR #556, opened 2023-02-13
Fix: Tailwind content path should match TS

joostmeijles:
### Description

Tailwind configuration is not working because the package content path does not on Typescript extensions.

### Type of Change


lfades (approved):
Thank you!

PR #557, opened 2023-02-15
Solving Deployment Issues with Django Vercel

alisolanki:
### Description

I've added Static files, build_files.sh and vercel.json file which will help in direct deployment of the example project without any additional settings.

### Demo URL

I had made a similar deployment using this repo: https://emotion-ai.alisolanki.com

### Type of Change

- [❎] New Example
- [✅] Example updates (Bug fixes, new features, etc.)
- [❎] Other (changes to the codebase, but not to examples)

### New Example Checklist

- [✅] 🛫 `npm run new-example` was used to create the example
- [✅] 📚 The template wasn't used but I carefully read the [Adding a new example](https://github.com/vercel/examples#adding-a-new-example) steps and implemented them in the example
- [✅] 📱 Is it responsive? Are mobile and tablets considered?

### DIFF
--- a/python/django/build_files.sh
+++ b/python/django/build_files.sh
@@ -1,0 +1,4 @@
+echo " BUILD STARTED "
+python3.9 -m pip install -r requirements.txt
+python3.9 manage.py collectstatic --noinput --clear
+echo " BUILD ENDED "
--- a/python/django/vercel.json
+++ b/python/django/vercel.json
@@ -3,9 +3,20 @@
     {
       "src": "vercel_app/wsgi.py",
       "use": "@vercel/python"
-    }
+    },
+    {
+      "src": "build_files.sh",
+      "use": "@vercel/static-build",
+      "config": {
+        "distDir": "staticfiles_build"
+      }
+    },
   ],
   "routes": [
+    {
+      "src": "/static/(.*)",
+      "dest": "/static/$1"
+    },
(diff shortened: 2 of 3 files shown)


leerob:
Can you explain a bit more what the problem this solves is?

alisolanki:
> Can you explain a bit more what the problem this solves is?

It solves 2 major issues
1. Adds static files while deployment (production level) (ref: [Django Docs](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-STATIC_ROOT:~:text=for%20multiple%20sites.-,Static%20Files,-%C2%B6))
2. Installs all the Python packages from requirements.txt file in Vercel's environment

leerob:
I'm confused, because I thought those things were working today, but I could be wrong.

Aren't the `requirements.txt` files picked up normally?

alisolanki:
> I'm confused, because I thought those things were working today, but I could be wrong.
> 
> Aren't the `requirements.txt` files picked up normally?

Yes apologies, the requirements.txt files are getting picked up normally.

However I still have a problem with static, when the debug is turned to False, which is required when you deploy to a production environment like Vercel, we need to add the lines that I added in the commit.

build_files.sh line 2 is not needed for fetching requirements.txt but others are for static files and that's still needed I believe.

PR #558, opened 2023-02-17
Update middleware example

mattschwarz:
### Description

Add common exclusions the URL rewrite to maintain page rendering capabilities and asset paths even in maintenance mode.
Referencing this discussion: https://github.com/vercel/next.js/discussions/38615

### Type of Change

- [ ] New Example

### New Example Checklist

### DIFF
--- a/edge-middleware/maintenance-page/middleware.ts
+++ b/edge-middleware/maintenance-page/middleware.ts
@@ -14,9 +14,24 @@ export async function middleware(req: NextRequest) {
   try {
     // Check whether the maintenance page should be shown
     const isInMaintenanceMode = await get<boolean>('isInMaintenanceMode')
+    
 
     // If is in maintenance mode, point the url pathname to the maintenance page
     if (isInMaintenanceMode) {
+      // Optional
+      // Exclude core Next.js files, static files, and files in the public folder
+      // that might affect rendering the maintenance page
+      const pathname = req.nextUrl.pathname
+      const PUBLIC_FILE = /\.(.*)$/
+      
+      if (
+        pathname.startsWith("/_next") || // exclude Next.js internals
+        pathname.startsWith("/static") || // exclude static files
+        PUBLIC_FILE.test(pathname)  // exclude all files in the public folder
+      ) {
+        return NextResponse.next()
+      }
+      
       req.nextUrl.pathname = `/maintenance`
 
       // Rewrite to the url


dominiksipowicz:
As far as I'm concerned, this change might be redundant since the middleware function is already limited to running only on specific paths by the `config.matcher`. Are you trying to prevent people from blindly copying code into their middleware that will run on all traffic?

mattschwarz:
> As far as I'm concerned, this change might be redundant since the middleware function is already limited to running only on specific paths by the `config.matcher`. Are you trying to prevent people from blindly copying code into their middleware that will run on all traffic?

Given the context of this specific example, and if you were to follow this example as-is, you'd most likely find that any maintenance page you create (assuming it's using any referenced dependencies like components, assets, etc) would be blocked resulting in a broken page...including things some basic things that I'd assume many may not consider, like your favicon. Maybe this could be a readme update to the example instead, but I think it's worth calling out that code like this will block _all_ paths. 

I do also think there's value in showcasing that there are certain paths you may want to leave open to traffic even when a site is in maintenance mode. For instance, paths your hosting on your app to assets referenced externally may not be links you want to break when taking a site down via something as quick/easy as an external config.

dominiksipowicz:
I see. So it might be better to refactor this example to run on all paths and then exclude assets like you did.

WDYT @lfades @goncy ?

lfades:
In the current rewrite all static assets will also get rewritten to the assets of the maintenance page, which is intended behavior.  It can't just be the browser URL and it won't include assets like those in the public folder unless this line explicitly includes them:

```ts
const isInMaintenanceMode = await get<boolean>('isInMaintenanceMode')
```

It currently does and we should change that, but once that's changed the behavior of also rewriting the static files should persist. @goncy can you check this one out as I might be missing something 🙏

goncy:
Hey @mattschwarz thanks for the PR! I get your point that sometimes when you go into maintenance mode, some "parts" of your app, like assets or static files can still be accessed. In this case we are showing how to get just one part of your app in maintenance mode, but I think we can add a section to the example saying that if you want to put your entire app in maintenance mode you should take care of excluding the assets and internal files from middleware. In that case we can leave a minimal example working but still covering your case, what do you think?

mattschwarz:
> Hey @mattschwarz thanks for the PR! I get your point that sometimes when you go into maintenance mode, some "parts" of your app, like assets or static files can still be accessed. In this case we are showing how to get just one part of your app in maintenance mode, but I think we can add a section to the example saying that if you want to put your entire app in maintenance mode you should take care of excluding the assets and internal files from middleware. In that case we can leave a minimal example working but still covering your case, what do you think?

Yeah I don't necessarily think we need to include my code change since the current example specifies a config matcher, but considering that this is the only example offered of implementing some kind of maintenance mode, I think it's a reasonable assumption that most would benefit from instruction on how to do that for an entire app. With that in mind, mentioning how and what paths to avoid accidentally targeting to keep critical links and functionality working would be nice.

PR #560, opened 2023-02-21
Added cron example

steven-tey:
### Description

Adds a cron example that retrieves the newest post from Hacker News at different intervals.

### Demo URL

https://cron-template.vercel.app/

### Type of Change

- [ ] New Example

### New Example Checklist


solutions/cron/README.md
```diff
@@ -0,0 +1,46 @@
+---
+marketplace: false
```
leerob, line 2: This needs updated

solutions/cron/README.md
```diff
@@ -0,0 +1,46 @@
...
+reason: https://github.com/lfades/static-tweet is very similar to this one
+---
+
+# Vercel Cron Job Example
+
+This example shows you how you can use Vercel Cron Jobs to update data at different intervals
+
+## Demo
+
+https://cron.vercel.app/
```
leerob, line 12: Update to correct URL

solutions/cron/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@vercel/analytics": "^0.1.10",
+    "@vercel/edge-config": "^0.1.1",
+    "@vercel/examples-ui": "^1.0.4",
+    "ms": "^2.1.3",
+    "next": "canary",
```
leerob, line 16: Do we need canary or just latest?

solutions/cron/vercel.json
```diff
@@ -0,0 +1,24 @@
+{
+  "crons": [
+    {
+      "path": "/api/cron/1m",
```
leerob, line 4: On the demo it looks like every minute and 10 minutes had the same story, is that a fluke?

solutions/cron/README.md
```diff
@@ -0,0 +1,71 @@
+---
+name: Vercel Cron Job Example
+slug: vercel-cron
+description: A Next.js app that uses Vercel Cron Jobs to update data at different intervals.
+framework: Next.js
+useCase:
+  - Cron
```
manovotny, line 7: Has this use case been created in Contentful already?

solutions/cron/pages/_app.tsx
```diff
@@ -0,0 +1,23 @@
...
+      path="solutions/cron"
+      description="How to use Vercel Cron Jobs to update data at different intervals"
+    >
+      <Component {...pageProps} />
+      <Analytics />
+    </Layout>
+  )
+}
+
+export default App
```
manovotny, line 23: Is there a reason we're use `pages` instead of `app`?

solutions/cron/vercel.json
manovotny: I thought this too until I hit refresh and then it diverged. Problem with `getStaticProps`?

manovotny:
I don't think the `github.svg` and `vercel.svg` are used. If they're not, can they be removed?

solutions/cron/pages/_app.tsx
steven-tey: A few reasons:
- `app` is still in beta
- folks are still familiar with `pages` 
- this template is a focus on Vercel Cron

If we update this to use `app` dir we'll have to update all the examples in here to use `app` dir – I believe that should be part of a larger project

solutions/cron/README.md
steven-tey: updated!

solutions/cron/README.md
steven-tey: updated!

solutions/cron/package.json
steven-tey: that's what we're using for other examples too so I just copied and pasted

solutions/cron/vercel.json
steven-tey: It can happen when the 1min and 10min cron intersect, but usually the 10min one should lag a bit behind the 1min one:

CleanShot 2023-02-21 at 17 56 14

solutions/cron/README.md
steven-tey: Yessir!! https://vercel.com/templates/cron
CleanShot 2023-02-21 at 17 56 59

solutions/cron/vercel.json
steven-tey: @manovotny oh good catch, I might wanna do res.revalidate instead...

steven-tey: refactored to using SWR to make sure data is always up to date <picture data-single-emoji=":checked:" title=":checked:">:checked:</picture>

solutions/cron/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <Text>
+          Each of the following sections are the{' '}
+          <Link
+            href="https://github.com/HackerNews/API#new-top-and-best-stories"
+            target="_blank"
+            rel="noreferrer noopener"
+          >
+            newest stories on Hacker News
+          </Link>{' '}
+          retrieved at a different intervals using{' '}
```
SuttonJack, line 64:
```suggestion
          retrieved at different intervals using{' '}
```

SuttonJack (approved):
lgtm

solutions/cron/pages/_app.tsx
leerob: I'm okay with keeping pages and trying to move many of these at once 👍

solutions/node-hello-world/README.md
```diff
@@ -0,0 +1,35 @@
+# Node Hello World
```
leerob, line 1:
```suggestion
# Node.js Hello World
```

solutions/node-hello-world/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Super simple Node.js + Vercel example that returns a "Hello World" response.
```
leerob, line 3:
```suggestion
Simple Node.js + Vercel example that returns a "Hello World" response.
```

solutions/cron/pages/index.tsx
steven-tey: Good catch, thank you!

leerob (approved):

solutions/cron/pages/api/cron/[cron].ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+}
+
+async function update(interval: string) {
+  const topstories = await fetch(
+    'https://hacker-news.firebaseio.com/v0/newstories.json?print=pretty'
+  ).then((res) => res.json())
+
+  const response = await fetch(
+    `https://api.vercel.com/v1/edge-config/${process.env.EDGE_CONFIG_ID}/items${
```
Comment (dferber90) [inline solutions/cron/pages/api/cron/[cron].ts:22]: We could get rid of the EDGE_CONFIG_ID function by using the parseConnectionString(process.env.EDGE_CONFIG) method. That way we only have a single env var that we need to set up. This also then allows us to set up the edge config via the deploy button

solutions/cron/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
+import { Layout, Text, Code, Page, Link } from '@vercel/examples-ui'
+import Post from '@/components/post'
+import Head from 'next/head'
+
+const intervals = [
+  {
+    id: '1m',
```
dferber90, line 7: This will hit the Edge Config rate limit

solutions/cron/vercel.json
dferber90: This likely happened as the daily write limit was reached. Daily limit is 480, but writing every minute leads to 1440 writes.

We shouldn’t use such a low number, as the rate limit is per account so this also breaks other edge config demos once the limit is reached

solutions/node-hello-world/README.md
```diff
@@ -0,0 +1,31 @@
+# Node.js Hello World
```
dferber90, line 1: Just double checking: Is this here on purpose? Didn’t see it mentioned in the pr description

solutions/cron/pages/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  {
+    id: '1d',
+    name: 'Every Day',
+    cron: '0 0 * * *',
+  },
+]
+
+const image = 'https://cron-template.vercel.app/thumbnail.png'
+
+export default function Home({ data }: { data: any }) {
```
dferber90, line 35: Could we render a warning if we detect that EDGE_CONFIG is not defined? You can likely copy it from the feature-flags-apple-store or maintenance-page examples.

PR #247, opened 2022-05-04
basic-password | Remove the use of Buffer module for atob

louisgv:
Per the edge doc, Buffer is not included in the runtime: https://nextjs.org/docs/api-reference/edge-runtime

### Description

The latest next version as of this writing will not allow global.Buffer to be used in the edge function. This work replace Buffer with atob call for the decoding of the basic auth data.

### Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New Example
- [ ] New feature in existing example

### New Example Checklist

- [ ] 🚀 Link to deployment URL on Vercel
- [ ] 📱 Is it responsive? Are mobile and tablets considered?
- [ ] 📚 `README.md` preview link in PR description (branch)
- [ ] ⚙️ Secrets have instructions for how to set up in the readme
- [ ] 🛫 `npm run new-example` was run to create the example

--[/U64eCftcNBXxHZRu3](https://www.plasmo.com/engineering/log/2022.05#update-2022.05.04)

PR #570, opened 2023-02-24
basic-auth-password | Remove the use of atob

3ru:
### Description

This sample has been modified by #247 to use `atob`, but the function is now flagged as [`deprecated`](https://github.com/microsoft/TypeScript/issues/45566).

|IDE|
|:--:|
|image|

Also, in the current Next, `ArrayBuffer` works with edge runtime, so reverting to the original works fine.
edge doc: https://nextjs.org/docs/api-reference/edge-runtime#web-standard-apis

Since this sample is often referenced, I thought it would be better to modify it rather than wait.

### Type of Change

- [ ] New Example

### New Example Checklist

### DIFF
--- a/edge-middleware/basic-auth-password/middleware.ts
+++ b/edge-middleware/basic-auth-password/middleware.ts
@@ -10,7 +10,7 @@ export function middleware(req: NextRequest) {
 
   if (basicAuth) {
     const authValue = basicAuth.split(' ')[1]
-    const [user, pwd] = atob(authValue).split(':')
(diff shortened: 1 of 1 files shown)


lfades (commented):
`Buffer.from` is a Node.js API, `ArrayBuffer` is not being used here.

3ru:
I apologize for the confusion caused by the misunderstanding.

3ru:
@lfades 
Since the `Buffer` yesterday is now supported, can we leave from `atob`? (I already deleted the forks😖).

PR #595, opened 2023-03-15
Add turbo to all examples

lfades:
### Description

This PR is about adding turborepo to every example so that our previews can all be cached and we use less compute with every PR to this repo. It's also a good default to have Turbo always there!

### Type of Change

- [ ] New Example

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,3 +1,5 @@
+eval "$(fnm env --use-on-cd)"
+
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
--- a/app-directory/css-in-js/package.json
+++ b/app-directory/css-in-js/package.json
@@ -9,9 +9,9 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@chakra-ui/react": "^2.4.6",
-    "@emotion/react": "^11.10.5",
-    "@emotion/styled": "^11.10.5",
+    "@chakra-ui/react": "^2.5.1",
+    "@emotion/react": "^11.10.6",
+    "@emotion/styled": "^11.10.6",
     "@vercel/examples-ui": "latest",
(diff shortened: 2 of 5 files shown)


.husky/pre-commit
```diff
@@ -1,3 +1,5 @@
+eval "$(fnm env --use-on-cd)"
```
mehulkar, line 1: not sure if this is the type of repo where this matters, but i'd generally put this in a separate commit/PR

app-directory/css-in-js/package.json
```diff
@@ -9,23 +10,24 @@
...
-    "autoprefixer": "^10.4.7",
+    "autoprefixer": "^10.4.14",
     "eslint": "latest",
     "eslint-config-next": "canary",
-    "postcss": "^8.4.14",
-    "tailwindcss": "^3.0.24",
-    "typescript": "^4.7.3"
+    "postcss": "^8.4.21",
+    "tailwindcss": "^3.2.7",
+    "turbo": "^1.8.3",
```
mehulkar, line 30: Do we have renovate/dependabot on these? Or should we use `latest` on this? cc @gsoltis on whether this repo should be part of our testing scenarios for releases

app-directory/i18n/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "build": {
+      "outputs": [".next/**"]
```
mehulkar, line 5:
```suggestion
      "outputs": [".next/**", "!.next/cache/**"],
```

mehulkar: We need to merge this PR https://github.com/vercel/turbo/pull/4159 cc @tknickman

mehulkar (approved):

PR #600, opened 2023-03-19
Remove redundant `requestHeaders` variable in chatgpt api example

0xturner:
### Description

Noticed that there is a `requestHeaders` object that is declared, however, it never actually gets used in this file. This example can be simplified by removing these redundant lines.

I tested with my personal OpenAPI key and the requests were still fulfilled as expected. 

The other alternative to removing the variable would be to pass `requestHeaders` as an additional parameter to `OpenAIStream` and allowing the default request headers to be overridden.

### Type of Change

- [ ] New Example

### DIFF
--- a/solutions/ai-chatgpt/pages/api/chat.ts
+++ b/solutions/ai-chatgpt/pages/api/chat.ts
@@ -28,15 +28,6 @@ const handler = async (req: Request): Promise<Response> => {
   ]
   messages.push(...body?.messages)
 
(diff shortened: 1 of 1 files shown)


juancamiloqhz:
I came here to do the same thing. The variable is declared in /pages/api/chat.ts but not actually used. It's declared in /utils/OpenAIStreams.ts

dominiksipowicz (approved):
great find! thank you!

PR #634, opened 2023-04-20
Update testing example to add Visual Testing and Storybook

lfades:
### Description

- Added Applitools Eyes for visual testing
- Added Storybooks
- Fixed Storybooks setup in microfrontends example

Todo:
- [x] Add more visual tests to the setup (I reached the free limit in Applitools atm)
- [x] Test the batch id in Applitools so that all tests results end up in the same batch
- [x] Update the readme with docs for storybook and applitools setup.

### Type of Change

- [ ] New Example


lpalmes (approved):
Looks good to me!

We talked offline about the failing builds (like microfrontends) but those are unrelated to this change. So it should be resolved in a followup

PR #681, opened 2023-05-03
fix: kysely seed script is broken because `current_timestamp` is passed as a string literal.

igalklebanov:
### Description

Kysely seed script is broken because I passed `current_timestamp` as a string literal instead of raw sql (using sql template tag).

Also forgot to add `if not exists`.

So sorry. 😞 

I'm rushing this, didn't run this.

### Demo URL

Here's how its done in tests..
https://github.com/kysely-org/kysely/blob/master/test/node/src/schema.test.ts#L70-L72

here's a playground link to see it matches the raw sql before move to built-in schema module code.
https://kyse.link/?p=s&i=UVScTQg9PZr8zSbGM63J

### Type of Change

- [ ] New Example

### New Example Checklist


steven-tey:
Thank you so much for the PR!

PR #727, opened 2023-06-06
feat(microfrontends): use npx over pnpm dlx for turbo ignore

dominiksipowicz:
use the same turbo ignore as the deployment button


lpalmes:
lgmt!

lpalmes (approved):

PR #756, opened 2023-07-24
warn when integration has not synced

dferber90:
### Description

Adds a warning in case the user did not yet cause the Split integration to sync into an Edge Config. In that case the demo would not work anyhow, so it's better to show instructions.

image

### DIFF
--- a/edge-middleware/feature-flag-split/app/missing-split-item/page.tsx
+++ b/edge-middleware/feature-flag-split/app/missing-split-item/page.tsx
@@ -19,7 +19,7 @@ function ExclamantionTriangleIcon() {
   )
 }
 
-export default function MissingEdgeConfigDialog() {
+export default function MissingSplitItemDialog() {
   return (
     <div
       className="relative z-30"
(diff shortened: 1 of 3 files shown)


edge-middleware/feature-flag-split/middleware.ts
```diff
@@ -14,4 +15,9 @@ export async function middleware(req: NextRequest) {
     req.nextUrl.pathname = `/missing-edge-config`
     return NextResponse.rewrite(req.nextUrl)
   }
+
+  if (!(await has(process.env.EDGE_CONFIG_SPLIT_ITEM_KEY))) {
```
lpalmes, line 19: Love this kind of feedback

edge-middleware/feature-flag-split/app/missing-split-item/page.tsx
```diff
@@ -0,0 +1,125 @@
...
+                      </span>
+                      , as defined by the EDGE_CONFIG_SPLIT_ITEM_KEY environment
+                      variable. If it does not contain such a key, you can go to
+                      Split and make a change to a feature flag. This causes the
+                      integration to resync. You should then see the value in
+                      your Edge Config.
+                    </li>
+                    <li className="list-item list-disc">
+                      Pull your latest Environment Variables if you are
+                      developing locally
```
lpalmes, line 97: It would be good to have the command as a hint at the end, I think `vercel env pull`

lpalmes (approved):

PR #800, opened 2023-10-06
cron example: Update .env.example with valid environment variables for KV

yhterrance:
### Description

Update the name of the environment variables so that the `@vercel+kv` library can read them properly.

### Type of Change

Bug Fixes


yhterrance:
Closing — no longer working in this repo. Thanks!

PR #858, opened 2024-01-19
chore(ci): use corepack to pin pnpm version

styfle:
https://twitter.com/styfle/status/1748373316874731803

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "examples",
   "license": "MIT",
   "private": true,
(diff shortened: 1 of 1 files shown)


delbaoliveira (approved):

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "examples",
   "license": "MIT",
   "private": true,
+  "packageManager": "pnpm@8.14.1+sha256.2df78e65d433d7693b9d3fbdaf431b2d96bb4f96a2ffecd51a50efe16e50a6a8",
```
hardfist, line 5: You may need to specify engines.pnpm to do
 double-check in case someone forget enable corepack

styfle: @hardfist Good point! Added 👍

anonrig: Nice catch!

PR #871, opened 2024-02-13
Add Vercel Toolbar and Flags integration to Hypertune example

miraan:
Add Vercel Toolbar and Flags integration to Hypertune example


edge-middleware/feature-flag-hypertune/app/.well-known/vercel/flags/route.ts
```diff
@@ -0,0 +1,9 @@
+import { NextResponse } from 'next/server'
+import { encryptApiData } from '@vercel/flags'
+import { vercelFlagDefinitions } from '../../../../generated/generated'
+
+export async function GET() {
+  const apiData = { definitions: vercelFlagDefinitions }
```
dferber90, line 6: This should be updated to the new `verifyAccess` method that landed in the newer vercel/flags package

```
import { NextResponse } from "next/server";
import { verifyAccess, type ApiData } from "@vercel/flags";
import { vercelFlagDefinitions } from '../../../../generated/generated'

export async function GET() {
  const access = await verifyAccess(request.headers.get('Authorization'));
  if (!access) return NextResponse.json(null, { status: 401 });

  const apiData: ApiData = { definitions: vercelFlagDefinitions }

  return NextResponse.json<ApiData>(apiData);
}
```

AndyBitz (approved):
Current version works, so we'll merge it and adjust it with the most recent updates.

Thanks!

PR #896, opened 2024-04-05
DataDome edge-middleware - fix `searchParams` in payload

MickaelDatadome:
### Description

Fix the way the `searchParams` is passed in the payload in the DataDome edge-middleware

### Type of Change

- [Fix] - Payload content for `searchParams`


edge-middleware/bot-protection-datadome/lib/datadome.ts
```diff
@@ -43,7 +43,7 @@ export default async function datadome(req: NextRequest) {
     PostParamLen: req.headers.get('content-length'),
     Protocol: req.headers.get('x-forwarded-proto'),
     Referer: req.headers.get('referer'),
-    Request: pathname + encode(Object.fromEntries(req.nextUrl.searchParams)),
+    Request: pathname + req.nextUrl.search,
```
amarlot-datadome, line 46: I see you removed the encode. It is not useful ? (Open question. It is still unclear for me when to use it and not)

MickaelDatadome: Yes, the method `encode` has been removed as it was used only for this specific part.
The encoding of the payload is done once, when calling the method `stringify` just before sending the payload to the API server. (line 72)

PR #907, opened 2024-04-23
CRON example: Update link to kv marketing page

JohnGemstone:
Current link to https://vercel.com/kv is blank and doesn't forward anywhere, updating to actual canonical marketing page https://vercel.com/storage/kv

### DIFF
--- a/solutions/cron/pages/index.tsx
+++ b/solutions/cron/pages/index.tsx
@@ -81,7 +81,7 @@ export default function Home({ data }: { data: any }) {
           </Link>{' '}
           and stored in{' '}
           <Link
-            href="https://vercel.com/kv"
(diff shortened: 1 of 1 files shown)


leerob:
Can you share more about what you see? That link redirects for me.

JohnGemstone:
> Can you share more about what you see? That link redirects for me.

Seems to be working ok for me too! Basically the page loaded onto a blank screen, no 404 or 500. I examined the network tab and didn't see any errors so I just assumed it was an old route and didn't investigate too much sorry.

This happened from around 17:56 GMT+1 on the 23rd April if you wanted to dig further on your logs. 

I'll get this closed since the redirects are working again.

Cheers!

PR #911, opened 2024-05-08
fix statsig usage

dferber90:
Prevent statsig from fetching in middleware.

- Turning on `localMode` will log an error on any network activity
- Setting `initStrategyForIDLists: "never"` will prevent Statsig from attempting to fetch the ID List over the network, as ID Lists are not supported by the Edge Config Adapter (yet)


AndyBitz (approved):

Lostthought1 (commented):
👍

PR #918, opened 2024-05-22
Fix pgvector example + upgrade

leerob:


peguerosdc:
Hi! Is there any specific reason why the rate limit was removed? Should I still use that implementation as a reference for my project?

leerob:
👋 Vercel has built-in rate limiting coming to our Firewall if you want this https://www.youtube.com/watch?v=t-U6vTrx_9Y

PR #924, opened 2024-06-05
Update Hypertune example to use the new Feature Flags pattern

miraan:
Update Hypertune example to use the new Feature Flags pattern


edge-middleware/feature-flag-hypertune/.env
```diff
@@ -0,0 +1,4 @@
+HYPERTUNE_FRAMEWORK=nextApp
+HYPERTUNE_PLATFORM=vercel
+HYPERTUNE_OUTPUT_DIRECTORY_PATH=generated
+HYPERTUNE_GET_HYPERTUNE_IMPORT_PATH=../lib/getHypertune
```
dferber90, line 4: Is HYPERTUNE_ADMIN_TOKEN missing here?

edge-middleware/feature-flag-hypertune/app/.well-known/vercel/flags/route.ts
dferber90, line 1: It looks like something is forcing the browser to refresh. And that seems to be mounting/unmounting the FlagValues, which then in turn triggers network request(s) to decrypt the flag values detected in the DOM.

This looks like a bug

https://github.com/vercel/examples/assets/1765075/0b075788-be51-450b-bb5c-64916d6aff70


_Not related to this file, but using this so we can have a threaded convesation_

edge-middleware/feature-flag-hypertune/app/page.tsx
```diff
@@ -26,13 +27,16 @@ export default async function Home() {
           with Vercel Edge Config to initialize the Hypertune SDK with near-zero
           latency on the server so you can access your feature flags and run A/B
           tests with no performance impact to your app. It also shows how to
-          integrate with the Vercel Toolbar to let you easily set local flag
-          overrides while developing.
+          integrate with the Vercel Toolbar to let you easily view and override
+          your feature flags without leaving your frontend. Finally, it shows
+          how to use the Vercel Feature Flags pattern to use flags on a static
```
dferber90, line 32: To clarify a bit: we consider the ability to define flags in code as our feature flags pattern, and the ability to precompute some flags and use middleware to route between static pages is basically just an extension of that.

dferber90 (changes requested):
Looks super promising!

edge-middleware/feature-flag-hypertune/.env
miraan: The following environment variables are set via the deploy flow: NEXT_PUBLIC_HYPERTUNE_TOKEN, HYPERTUNE_ADMIN_TOKEN, EDGE_CONFIG, EDGE_CONFIG_HYPERTUNE_ITEM_KEY, FLAGS_SECRET

They are needed in production. The instructions say to pull these to .env.development.local to run the project locally.

The environment variables here in .env are not needed in production. They are used by the CLI tool for codegen only.

edge-middleware/feature-flag-hypertune/app/page.tsx
miraan: We show the middleware + routing between static pages in this example so thought I'd call it out. But will update the wording here!

dferber90 (approved):

PR #932, opened 2024-07-01
Toolbar Feature Flags SvelteKit example

dummdidumm:
### Description

Simple SvelteKit template that uses the [Vercel Toolbar](https://vercel.com/docs/workflow-collaboration/vercel-toolbar) and [Feature Flags](https://vercel.com/docs/workflow-collaboration/feature-flags).

### Demo URL

https://toolbar-feature-flags-sveltekit.vercel.app/

### Type of Change

- [x] New Example

### New Example Checklist


toolbar/toolbar-feature-flags-sveltekit/src/routes/+page.server.ts
```diff
@@ -0,0 +1,8 @@
+import { shouldUseSvelteColor } from '$lib/flags'
+
+export async function load() {
+  return {
+    // Requests the feature flag and returns its value
+    useSvelteColor: await shouldUseSvelteColor(),
```
dferber90, line 6: Will people think about hooks when we prefix this with "use"?

toolbar/toolbar-feature-flags-sveltekit/src/lib/flags.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const shouldUseSvelteColor = flag<boolean>({
+  key: 'useSvelteColor',
+  description: 'Use the Svelte color for the headline', // optional
+  origin: 'https://example.com/#useSvelteColor', // optional
+  options: [{ value: true }, { value: false }], // optional
+  // can be async and has access to the event
+  decide(event) {
+    // Very simple example to show how to use the URL to determine the flag.
+    // In real live you would probably query your database or a service.
```
dferber90, line 11: Let's mention Edge Config here as it's our purpose built data store for feature flags, which will have better latency

toolbar/toolbar-feature-flags-sveltekit/src/routes/+page.server.ts
```diff
@@ -0,0 +1,8 @@
...
+  }
+}
```
dferber90, line 8: Have you tested this without the `FLAGS_SECRET` env var locally? I expect people will clone this without pulling env vars, and it would be great to show some instructions within the app in case the env var is missing

dummdidumm: The README has a step by step guide what you should do to clone and run it locally, which includes the generation + pulling of the env vars

toolbar/toolbar-feature-flags-sveltekit/src/routes/+page.server.ts
dummdidumm: React-heads might yeah - in fact eslint did error when I called the flag `useSvelteColor` 😆 I'll come up with a different name

toolbar/toolbar-feature-flags-sveltekit/src/routes/+page.server.ts
dferber90: If it's not even building for dev without the secret we're good <picture data-single-emoji=":thumbs:" title=":thumbs:">:thumbs:</picture>

PR #935, opened 2024-07-09
Fix fallback prop to image src attribute

sarnakov:
### Description

The issue is that fallback prop is not being passed to the Image component.

### Type of Change

- [ ] Bug fixes


mar-js (approved):

leerob:
It appears we're doing `fallback = fallbackImage`, so either of the values should work here.

PR #977, opened 2024-11-14
Remove `express` install and use directions that don't work

trek:
I'm assuming these instructions worked at some point when the express example got shifted into a monorepo the updated instructions weren't tested. As written it is not possible to do this in a one-liner with `git`

The closes think we could come up with was (h/t @TooTallNate)

```
curl -sfLS https://github.com/vercel/examples/archive/refs/heads/main.tar.gz | tar -xv --strip-components=2 examples-main/solutions/express
```

But the general reaction was 🤢 .

I think we just need to remove this at the moment.


trek:
Closing this in favor of @yumin-jung's solution.

PR #1050, opened 2025-01-24
Update Blitz boilerplate to 2.x

trek:
Bumped to 2.x


framework-boilerplates/sveltekit/pnpm-lock.yaml
EndangeredMassa, line 1: **question (soft-blocking):** ‏Why was this file also changed?

trek: https://github.com/vercel/examples/pull/1051

PR #1139, opened 2025-05-22
feat: add posthog adapter example

joshsny:
### Description

Add an example for using the `@flags-sdk/posthog` adapter.

### Type of Change

- [ ] New Example

### New Example Checklist


flags-sdk/posthog/.env.example
```diff
@@ -0,0 +1,9 @@
+# For @flags-sdk/posthog
+FLAGS_SECRET=""
```
joshsny, line 2: @AAorris should the user always generate this with the toolbar? I used `openssl rand -hex 32` but that caused me 401s when I ended up using the toolbar.

flags-sdk/posthog/app/opengraph-image.jpg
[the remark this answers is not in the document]
AAorris: Flags SDK + PostHog - Light

flags-sdk/posthog/flags.ts
```diff
@@ -0,0 +1,32 @@
...
+  postHogKey: process.env.NEXT_PUBLIC_POSTHOG_KEY as string,
+  postHogOptions: {
+    host: process.env.NEXT_PUBLIC_POSTHOG_HOST as string,
+
+  },
+})
+
+
+
+export const showSummerBannerFlag = flag<boolean, PostHogEntities>({
```
AAorris, line 15: It would be good to recommend the default adapter — If there are defaults that you don't recommend there, we should integrate that feedback!

```suggestion
import { postHogAdapter, type PostHogEntities } from '@flags-sdk/posthog'
import { flag } from 'flags/next'
import { identify } from './lib/identify'

export const showSummerBannerFlag = flag<boolean, PostHogEntities>({
```

flags-sdk/posthog/app/.well-known/vercel/flags/route.ts
```diff
@@ -0,0 +1,40 @@
+import { getProviderData, createFlagsDiscoveryEndpoint } from 'flags/next'
+import { getProviderData as getPostHogProviderData } from '@flags-sdk/posthog'
+import { mergeProviderData } from 'flags'
+import * as flags from '../../../../flags'
+
+export const dynamic = 'force-dynamic' // defaults to auto
+
+export const getAppHost = () => {
```
AAorris, line 8: Is this consistent enough that we should have this in the adapter itself?

AAorris (approved):
We definitely need to update the opengraph image but the rest of my feedback is non blocking. While I suggested an OG image you can bring your own for the example if desired.

flags-sdk/posthog/app/.well-known/vercel/flags/route.ts
joshsny: It is - this should be moved to the adapter 👍

AAorris: Moved to the adapter and made `host` optional so you can remove it (or pass a resolved `appHost`) — version 0.1.0-2e7a523a-20250522190315

flags-sdk/posthog/app/opengraph-image.jpg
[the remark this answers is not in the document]
joshsny: updated!

AAorris:
> @AAorris do you mind taking another look here to see if we are happy?

@joshsny I'll take another look this morning 👍

It would be good to remove the `getAppHost` function from the well known route now that the adapter gets a default host using this implementation

joshsny:
> It would be good to remove the getAppHost function from the well known route now that the adapter gets a default host using this implementation

That would be a good idea wouldn't it 🤦

Will update now!

AAorris (approved):

PR #1164, opened 2025-07-29
updated nuxt example to v4

Dino-Kupinic:
### Description

Updated nuxt example to v4

### Type of Change

- [ ] New Example


Dino-Kupinic:
> Should we rename the dir to `nuxt/` instead of `nuxtjs/`?

Renamed it, anything else?

PR #1211, opened 2025-08-23
[vibe-coding-platform] Ship latest updates

javivelasco:
This PR brings a number of improvements to the `vibe-coding-platform` example:

- Fixed a bug persisting the prompt in local storage.
- Moved from two tools (run command + wait command) to a single tool that takes a wait parameter to distinguish if the LLM needs to wait for completion or if the command is meant to run in background. This makes the LLM more reliable.
- Moved from a sub-call to the LLM that generates paths & files to one where the main LLM call decides on the paths and requests just the content.
  - This allows for the main LLM call to have better context on the files that were already generated.
  - It reduces a lot hallucinations where it tries to upload the same files multiple times.
- Integrated some ai-elements:
  - The conversation (including auto-scroll) is now from our newly component lib.
  - The spinners we show are also from the library.
- Included error handling for all the tool call steps that are possible to fail.
  - This caused some issues where tool calls where shown in an inconsistent state, now they reflect the error status.
  - This also allows to return information for the LLM to fix in the next iteration instead of swallowing the errors.
  - It makes the whole workflow more reliable preventing the effect of hanging where actually there was a failure.
- Introduced a self-correction feature that triggers once to prevent infinite iterations if the system is not capable to get to a solution.
- Introduced settings to disable auto-correct and reasoning level for GPT-5.

### DIFF
--- a/apps/vibe-coding-platform/app/chat.tsx
+++ b/apps/vibe-coding-platform/app/chat.tsx
@@ -12,7 +12,6 @@ import {
 import { Input } from '@/components/ui/input'
 import { Message } from '@/components/chat/message'
 import { ModelSelector } from '@/components/settings/model-selector'
-import { MoonLoader } from 'react-spinners'
 import { Panel, PanelHeader } from '@/components/panels/panels'
 import { Settings } from '@/components/settings/settings'
 import { useChat } from '@ai-sdk/react'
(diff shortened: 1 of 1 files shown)


brrock:
MCP?

PR #1474, opened 2026-05-05
docs: clarify hono mcp setup

### LINKED ISSUE #1217: hono-mcp example is missing steps/details
Cloning https://github.com/vercel/examples.git and switching to starter/hono-mcp does not work. 

Tried `npm i` and then `vercel dev` with defaults from the vercel cli and pointed to ./src 

Running `vc dev` afterwards and visiting http://localhost:3000 returns the index.tsx as text

Moving indes from ./src to / makes it run.

EfeDurmaz16:
## Summary
- Clarifies that local commands should be run from `starter/hono-mcp` rather than `src`
- Expands setup details for the Vercel CLI project root prompt
- Uses explicit `vercel` commands in the README examples

Fixes #1217.

## Verification
- git diff --check

### DIFF
--- a/cdn/mintlify-docs-rewrite/vercel.ts
+++ b/cdn/mintlify-docs-rewrite/vercel.ts
@@ -1,6 +1,7 @@
 import { routes, type VercelConfig } from '@vercel/config/v1'
 
-const MINTLIFY_DOCS_URL = process.env.MINTLIFY_DOCS_URL || 'https://vercel-fcadfe60.mintlify.dev'
+const MINTLIFY_DOCS_URL =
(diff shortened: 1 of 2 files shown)


EfeDurmaz16:
Updated this branch after checking the Vercel schema failure.

Current branch state:
- `cdn/mintlify-docs-rewrite/vercel.ts` uses `routes.rewrite()` for both `/docs` and `/docs/:match*`, so the generated rewrite entries include destinations instead of the previous invalid shape.
- Added a small formatting-only follow-up for that config file.

Verification:
- `pnpm install --frozen-lockfile`
- `pnpm exec prettier --check cdn/mintlify-docs-rewrite/vercel.ts`
- static check confirmed both `routes.rewrite(...)` entries are present

Commits: `a8497a59`, `d5527431`

PR #1493, opened 2026-06-11
node server.ts example

ricardo-agz:
### Description

Starter example for deploying a bare Node server with `server.ts`

### Demo URL

https://vercel-node-serverts.labs.vercel.dev/

### Type of Change

- [x] New Example

### New Example Checklist

### DIFF
--- a/starter/node-server/src/pages/home.ts
+++ b/starter/node-server/src/pages/home.ts
@@ -1,0 +1,85 @@
+export const homePage = `
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>server.ts</title>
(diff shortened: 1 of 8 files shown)


starter/node-server/src/pages/home.ts
```diff
@@ -0,0 +1,85 @@
+export const homePage = `
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>server.ts</title>
+    <link rel="icon" type="image/svg+xml" href="/favicon.ico">
```
jeffsee55, line 8: Where would this come from? Would need a `public` folder, we should test that works.

jeffsee55 (approved):
Maybe we should throw a couple of assets like css in `public` to make sure that works.

PR #1519, opened 2026-06-26
Add a flask ai chat example.

gscho:
### Description

Adds a flask version of https://github.com/vercel/examples/tree/main/websockets/fastapi-ai-chat

**This needs to be released before merging:** https://github.com/vercel/vercel/pull/16811

### Demo URL

Requires an AI gateway key so you need to deploy it yourself to test.

### Type of Change

- [ ] New Example

### New Example Checklist


websockets/flask-ai-chat/backend/main.py
```diff
@@ -0,0 +1,58 @@
...
+        async for event in stream:
+            if isinstance(event, ai.events.TextDelta):
+                ws.send(json.dumps({"type": "text_delta", "content": event.chunk}))
+    ws.send(json.dumps({"type": "text_done"}))
+
+
+@sock.route("/api/ws")
+def websocket_endpoint(ws):
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
```
fantix: Would `asyncio.run()` work here instead of manually coping with event loops?

websockets/flask-ai-chat/backend/main.py
```diff
@@ -0,0 +1,58 @@
+import asyncio
+import json
+import os
+
+import ai
+from flask import Flask
+from flask_sock import Sock
```
fantix, line 7: I'm seeing many Flask WebSocket tutorials using `flask_socketio` (from the same author but with a lot more GH :star:) - my personal wish here is that we have a SocketIO Python example too (also being slightly different than the existing one) but that doesn't have to be this PR.

fantix (approved):
Looks good to me!

> Requires an AI gateway key so you need to deploy it yourself to test.

Maybe we can try https://github.com/vercel-labs/ai-python/pull/183?

websockets/flask-ai-chat/backend/main.py
gscho: I tested a version using that library and it worked as well. We could do a new example for that one

PR #1541, opened 2026-07-28
Add starter/unstructured-transform-ai-sdk

mikedownesdev:
Adds `starter/unstructured-transform-ai-sdk` — a Next.js chat app that connects the [AI SDK](https://ai-sdk.dev) to [Unstructured's Transform MCP server](https://docs.unstructured.io/transform/overview).

Paste a public document URL and the model parses it (PDF, DOCX, XLSX, images, and 70+ formats) into clean Markdown via MCP tool calls — no parsing code of your own. The model drives the full async job lifecycle: start the transform, pace status polling with a `wait` helper, fetch results, and download the parsed Markdown.

## Included

- One-click **Deploy to Vercel** button with the two required env vars pre-declared.
- `.env.example` and setup instructions.
- MIT `LICENSE`.
- A keyless **preview mode**: deployed without keys, the app renders the UI and explains how to run it live, instead of erroring on the first request.

## Environment variables

- `ANTHROPIC_API_KEY` — the chat model (Claude).
- `UNSTRUCTURED_API_KEY` — authenticates the Transform MCP server ([get one](https://transform.unstructured.io)).

## Notes / verified

- `npm install`, `tsc --noEmit`, and `next build` all pass.
- The full parse flow was verified end-to-end against the production server (`https://mcp.transform.unstructured.io`).
- The `downloadText` tool is restricted to the Transform host over HTTPS to avoid SSRF from the serverless function.
- Opened as a **draft** — happy to adjust placement, naming, or add gallery frontmatter to match your submission conventions.

### DIFF
--- a/starter/unstructured-transform-ai-sdk/.env.example
+++ b/starter/unstructured-transform-ai-sdk/.env.example
@@ -1,0 +1,11 @@
+# Your Anthropic API key — powers the chat model (Claude).
+# Get one at https://console.anthropic.com/settings/keys
+ANTHROPIC_API_KEY=sk-ant-...
+
+# Your Unstructured Transform API key — authenticates the Transform MCP server.
+# Get one at https://transform.unstructured.io/get-started
+# NOTE: this is a Transform key, which is distinct from a Platform API key.
+UNSTRUCTURED_API_KEY=...
+
+# Optional: override the Transform MCP endpoint (defaults to production root).
+# UNSTRUCTURED_MCP_URL=https://mcp.transform.unstructured.io
--- a/starter/unstructured-transform-ai-sdk/.gitignore
+++ b/starter/unstructured-transform-ai-sdk/.gitignore
@@ -1,0 +1,23 @@
+/node_modules
+/.next
+/out
+/build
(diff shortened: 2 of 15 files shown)


mikedownesdev:
Thanks for flagging this. I checked it against the live server and the premise doesn't hold for this MCP server, so the allowlist is working as intended.

The concern assumes the pre-signed `download_url` is served from cloud object storage (a different host). For Unstructured's Transform server, the results tool returns a `download_url` **on the MCP host itself** (`https://mcp.transform.unstructured.io/output/…?token=…`), so it matches the allowlist and the download succeeds. Verified end-to-end against production:

```
download_url host: mcp.transform.unstructured.io
GET status: 200 | final host after redirects: mcp.transform.unstructured.io | bytes: 150883
```

The full agent flow also completes — the model downloads and summarizes the parsed Markdown. The host restriction is deliberate SSRF hardening (so a crafted prompt can't turn `downloadText` into a fetch of internal/metadata URLs), and since the real download URL is on the allowed host, it doesn't block legitimate downloads. Leaving it as-is.

PR #1544, opened 2026-07-30
Revert Global Config rename for LaunchDarkly examples

AndyBitz:
Reverts the LaunchDarkly templates to the previous Edge Config SDK until their package can be updated to fix a type issue that came up between `@vercel/edge-config@1.2.0` and `vercel/edge-config@1.1.1`

### DIFF
--- a/edge-middleware/feature-flag-launchdarkly/README.md
+++ b/edge-middleware/feature-flag-launchdarkly/README.md
@@ -1,7 +1,7 @@
 ---
 name: LaunchDarkly Integration example
 slug: feature-flag-launchdarkly
-description: Learn how to set up the LaunchDarkly integration to read flags from Global Config
+description: Learn how to set up the LaunchDarkly integration to read flags from Edge Config
 framework: Next.js
 useCase: Edge Middleware
 css: Tailwind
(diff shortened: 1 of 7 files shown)


lfades (approved):

joker23:
Hi @AndyBitz - was directed to this PR by a coworker in LD. I noticed that the `@launchdarkly/vercel-server-sdk` is locked at `1.3.3` in the lockfile (ref https://github.com/vercel/examples/pull/1544/changes#diff-78a9d7d39d607e32e820fb6c4ea11c5177964acd22296b13c48394568481eda5R11-R13). Could we have this updated to the latest (`1.3.53` published 2 days ago). That version should be built with `edge-config` at `1.4.x`.

Given that this is a bunch of patch version bumps, I think we will need to test to make sure everything is still operational afterwards :).

AndyBitz:
> Hi @AndyBitz - was directed to this PR by a coworker in LD. I noticed that the `@launchdarkly/vercel-server-sdk` is locked at `1.3.3` in the lockfile (ref https://github.com/vercel/examples/pull/1544/changes#diff-78a9d7d39d607e32e820fb6c4ea11c5177964acd22296b13c48394568481eda5R11-R13). Could we have this updated to the latest (`1.3.53` published 2 days ago). That version should be built with `edge-config` at `1.4.x`.
> 
> Given that this is a bunch of patch version bumps, I think we will need to test to make sure everything is still operational afterwards :).

@joker23 good idea! I oversaw that `@launchdarkly/vercel-server-sdk` [does not lock the version for](https://github.com/launchdarkly/js-core/blob/main/packages/sdk/vercel/package.json#L38) `@vercel/edge-config`.

I opened https://github.com/vercel/examples/pull/1546 which bumps `@launchdarkly/vercel-server-sdk` to the latest version, which in turn forces it to use `@vercel/edge-config@1.5.0`, which is compatible with `@vercel/global-config`.

PR #1550, opened 2026-08-06
bun(websockets): add bun websocket example

ecklf:
## TLDR

Adds a deployable native Bun WebSocket echo example for Vercel.

## Problem

The WebSocket examples do not show how to deploy a dependency-free Bun application using `Bun.serve()` and its native WebSocket handlers on Vercel.

## Solution

Adds a `websockets/bun` example with a native `/ws` upgrade handler, echo behavior, health endpoint, and responsive browser client. Includes Bun framework preset detection, pinned type dependencies, public registry lockfile entries, Vercel runtime configuration, explicit client asset bundling, local scripts, and one-click deployment instructions.

### DIFF
--- a/websockets/bun/vercel.json
+++ b/websockets/bun/vercel.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "bunVersion": "1.x",
+  "functions": {
+    "server.ts": {
+      "includeFiles": "public/index.html"
+    }
+  }
+}
--- a/websockets/bun/vercel.json
+++ b/websockets/bun/vercel.json
@@ -1,9 +1,4 @@
 {
   "$schema": "https://openapi.vercel.sh/vercel.json",
-  "bunVersion": "1.x",
-  "functions": {
-    "server.ts": {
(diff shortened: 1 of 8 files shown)


websockets/bun/vercel.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "bunVersion": "1.x",
+  "functions": {
```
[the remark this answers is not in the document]
ecklf: this deploys just fine, seems like false alarm

websockets/bun/vercel.json
```diff
@@ -0,0 +1,9 @@
...
+    "server.ts": {
+      "includeFiles": "public/index.html"
```
Schniz: Can we avoid this by making "/": import("./public/index.html")?

Schniz (approved):

websockets/bun/vercel.json
ecklf: probably we can drop this entirely, i just forgot

the main reason was I wanted to use `Bun.file` to showcase

PR #1553, opened 2026-08-07
fix: build errors for examples

ecklf:
### Description

Fixes build failures surfaced while validating the Bun WebSocket example PR:

- **Mintlify docs rewrite:** Vercel config validation reported rewrite entries without a `destination`. Use `routes.rewrite()` so dynamic destinations are emitted correctly.
- **Split feature flag example:** `@vercel/global-config` initialized during module evaluation and failed when no connection string was available during the build. Create the client lazily when handling a request.
- **Upstash redirects:** the documented `POPULATE_REDIS=false` opt-out was treated as truthy. Compare the value explicitly so population can be disabled.
- **Turborepo Hono API:** TypeScript could not resolve `@repo/constants` because the workspace dependency had not been built. Build it before compiling the API.
- **Blob SvelteKit:** Turbo restored `.svelte-kit` but omitted the adapter-generated `.vercel/output`, causing Vercel to report a missing output directory. Include `.vercel/output/**` in cached build outputs.

Vercel project root settings were also corrected for the Services examples; their redeployments now pass.

### Verification

- Ran the affected `pnpm` builds locally.
- Validated the Mintlify config with Vercel CLI 58.1.0.
- Confirmed passing Vercel deployments for Blob SvelteKit, Split, Mintlify, and both Hono API projects.

The Upstash deployment still depends on replacing or disabling its stale Redis endpoint.

### Type of Change

- [ ] New Example


tiago-loureiro (approved):

edge-middleware/redirects-upstash/scripts/upstash.js
```diff
@@ -61,7 +61,7 @@ async function setupUpstash() {
 
   await writeFile(filePath, content)
 
-  if (!process.env.POPULATE_REDIS) {
+  if (process.env.POPULATE_REDIS === 'false') {
```
kldavis4: this flips the default, not just the opt-out. before: unset or empty meant skip. now anything that isn't the exact string `false` populates, so an env without `POPULATE_REDIS` falls into the try block, hits `upstash()`, and throws `Missing required Upstash credentials of the REST API` at line 10. that build used to succeed with just the 1000 hardcoded redirects

probably fine inside this repo since `.env` is committed with `POPULATE_REDIS = true`, but anyone copying just the app dir gets a hard build failure where they didn't before. maybe:

```js
const populate = process.env.POPULATE_REDIS
if (populate === 'false' || !populate) {
```

also worth noting this doesn't actually fix the red check — `edge-functions-redirects-upstash` is still failing and the committed `.env` keeps it at `true`, so the new opt-out never gets exercised in CI. I think either set it to `false` in `.env` (or as a project env var, since next won't override process.env) or pull this hunk out into its own PR? otherwise it looks fixed but isn't

starter/turborepo-with-hono/apps/api/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "api",
   "type": "module",
   "scripts": {
-    "build": "tsc",
+    "build": "pnpm --filter @repo/constants build && tsc",
```
kldavis4: `turbo.json` already has `"build": { "dependsOn": ["^build"] }`, so `turbo run build` builds `@repo/constants` first. with this change it gets built twice on a cold build — once under turbo's cache via `^build`, then again as an uncached nested `pnpm --filter` inside `api#build`. harmless today since `apps/web` doesn't import constants, but it rewrites `packages/constants/dist` outside turbo's knowledge, which is a torn-output risk the moment a second parallel consumer shows up

I think the real issue is that the api project isn't building through turbo from the repo root — there's no `vercel.json` anywhere under `starter/turborepo-with-hono`, so it's relying on dashboard settings. `redirects-upstash` and `blob-sveltekit` both pin `"buildCommand": "pnpm turbo build"` in a checked-in `vercel.json`, seems like the same thing would work here and keeps the dependency expressed in one place

minor, but this also makes the api build pnpm-only while `apps/api/README.md` still tells people to run `npm install && npm run dev`

edge-middleware/redirects-upstash/scripts/upstash.js
ecklf: I restored population as explicit opt-in with `POPULATE_REDIS=true` and set the committed .env value to false. Imo it makes more sense to do the skip path by default?

starter/turborepo-with-hono/apps/api/package.json
ecklf: I've added a pin for the build command in `vercel.json` do you think that's the way to go here?

starter/turborepo-with-hono/vercel.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "buildCommand": "pnpm turbo build"
```
kldavis4: I don't think this file gets read - both api projects root at `starter/turborepo-with-hono/apps/api`, and vercel reads `vercel.json` from the root directory, so this is one level too high. wrote up the details in the thread on `apps/api/package.json`

net effect is `turborepo-hono-api` and `examples-q42o` both went red on this push, they were green before. so this needs to move to `apps/api/vercel.json` with a `--filter=api`, or the package.json change needs to come back

edge-middleware/redirects-upstash/.env
```diff
@@ -1 +1 @@
-POPULATE_REDIS = true
\ No newline at end of file
+POPULATE_REDIS = false
```
kldavis4, line 1: heads up on the side effect - with population off the build only writes the first 1000 redirects into `lib/redirects.json`, so `/1001` through `/10000` fall through to the `HGET` in `lib/redirects.ts`, miss, and 404. that's basically the demo's headline feature

the redis endpoint was already stale so it's not a new outage, but it does turn a loud build failure into a quietly broken demo. not blocking the merge imo, just think it's worth an issue to restore the backend so it doesn't sit like this

starter/personalization-builder-io/components/Link/Link.tsx
```diff
@@ -1,10 +1,10 @@
 import NextLink from 'next/link'
 
-export const Link: React.FC<React.AnchorHTMLAttributes<any>> = ({
+export const Link = ({
   href,
   children,
   ...props
-}) => {
+}: React.AnchorHTMLAttributes<HTMLAnchorElement>) => {
```
kldavis4, line 7: confirmed this one fixes the `renderLink` error - installed the project and ran tsc both ways. on main you get `Type 'ReactNode | Promise<ReactNode>' is not assignable to type 'ReactNode'` at `[[...path]].tsx:98`, with your version it's gone. tightening `any` -> `HTMLAnchorElement` is fine too since `renderLink`'s param is `AnchorHTMLAttributes<any>`, so the contravariant check still passes

but the project still doesn't build, there's a second error this doesn't touch:

```
./pages/_middleware.tsx:10:59
Type error: Argument of type 'RequestCookies' is not assignable to parameter of type 'Record<string, string>'.
  Index signature for type 'string' is missing in type 'RequestCookies'.
```

no `typescript.ignoreBuildErrors` in `next.config.js` so `next build` fails on it, which is why `starter-test` is still red

also that file is the next 12 nested-middleware convention. docs want it at the project root now ("Create a `proxy.ts` (or `.js`) file in the project root ... at the same level as `pages` or `app`") and there's no root `middleware.*` or `proxy.*` here, so on next 16 it isn't wired up as middleware at all - it's dead code that only exists to break the typecheck. probably wants deleting, or migrating to a root `proxy.ts` (`req.cookies` is a `RequestCookies` instance now, so `getPersonalizedRewrite` needs something like `Object.fromEntries(req.cookies.getAll().map(c => [c.name, c.value]))`)

edge-middleware/redirects-upstash/scripts/upstash.js
kldavis4: yeah agreed, skip by default is the right call. opt-in means a fresh clone with no upstash creds still builds, which is the first thing anyone hits. this is better than what I suggested

one thing though - `README.md:58` still says the opposite: "if you prefer not to do that then set `POPULATE_REDIS` to `false` in `.env`". that reads as populate-by-default with a `false` opt-out, so anyone following the readme gets no population and no hint why. probably wants a line saying set it to `true` to populate

also `.env.example` still doesn't mention `POPULATE_REDIS` at all, might be worth adding while you're in here

starter/turborepo-with-hono/apps/api/package.json
kldavis4: right idea but I think the placement is off, and it looks like it regressed the deploys - `turborepo-hono-api` and `examples-q42o` are both failing now and they were green on the last push

decoded the vercel bot metadata on this PR, both api projects have `rootDirectory` = `starter/turborepo-with-hono/apps/api`. vercel reads `vercel.json` from the root directory, so the new file at `starter/turborepo-with-hono/vercel.json` is a level above where either project looks and never gets read. and with the package.json change reverted, tsc can't resolve `@repo/constants` again, so we're back to the original error. `turborepo-hono-web` is still green just because web doesn't import constants

my earlier comment was underspecified here, sorry - the siblings I pointed at are single-app projects where the repo dir is the root dir, which isn't the case in a real monorepo. I think it wants to be:

```jsonc
// starter/turborepo-with-hono/apps/api/vercel.json
{
  "$schema": "https://openapi.vercel.sh/vercel.json",
  "buildCommand": "cd ../.. && pnpm turbo build --filter=api",
  "ignoreCommand": "pnpm dlx turbo-ignore"
}
```

the `--filter` matters here, bare `pnpm turbo build` would build web on every api deploy too

or honestly just restoring the package.json line is fine - it was green, and the double-build thing I flagged is a local dx wart, not a deploy failure

edge-middleware/redirects-upstash/.env
ecklf: The DB was broken to begin with. I've repopulated the DB with a new storage configuration and it seems now to work fine

edge-middleware/redirects-upstash/scripts/upstash.js
ecklf: Fixed, also added the example env

starter/personalization-builder-io/proxy.ts
```diff
@@ -0,0 +1,19 @@
+import { getPersonalizedRewrite } from '@builder.io/personalization-utils'
+import { type NextRequest, NextResponse } from 'next/server'
+
+const excludededPrefixes = ['/favicon', '/api']
+
+export function proxy(request: NextRequest) {
```
kldavis4, line 6: needs a `config.matcher`. `getPersonalizedRewrite` rewrites *any* pathname once the visitor has a `builder.userAttributes.*` cookie, and `excludededPrefixes` only covers `/favicon` and `/api`, so `_next/static` goes through it

same chunk url, with and without the cookie:

```
no cookie:   200  application/javascript  13809 bytes
with cookie: 200  text/html                1581 bytes
```

so every script and stylesheet breaks for personalized visitors. new behavior too - the old `pages/_middleware.tsx` was the next 12 convention, it never ran on 16

```ts
export const config = {
  matcher: ['/((?!api|_next/static|_next/image|favicon.ico).*)'],
}
```

think this is the only blocker, and CI won't catch it - build's green, the break is at runtime

starter/turborepo-with-hono/apps/api/vercel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "buildCommand": "cd ../.. && pnpm turbo build --filter=api",
```
kldavis4, line 3: works, but CI hasn't proven it - `turborepo-hono-api` and `examples-q42o` both say "Canceled by Ignored Build Step", so nothing built on this commit. the `ignoreCommand` I suggested is what's hiding it, sorry

ran it locally from `apps/api` and it's correct, constants builds then api. worth forcing one real build before merge though

nit: turborepo wants `turbo run build` when it's written into config or CI, bare `turbo build` is for interactive use. the sibling vercel.jsons do the same thing so no strong opinion

edge-middleware/redirects-upstash/lib/upstash.ts
```diff
@@ -35,22 +35,17 @@ export async function upstashRest(
   args: any[],
   options?: { pipeline: boolean }
 ) {
-  const domain = process.env.UPSTASH_REST_API_DOMAIN
-  const token = process.env.UPSTASH_REST_API_TOKEN
+  const url = process.env.KV_REST_API_URL
+  const token =
+    process.env.KV_REST_API_READ_ONLY_TOKEN ?? process.env.KV_REST_API_TOKEN
```
kldavis4, line 40: `||` rather than `??` here I think - a dashboard var left blank comes through as `""`, not undefined, so it won't fall back. token ends up `""` and you get `Missing required KV REST API credentials` instead of just using the write token

also the old protocol guard is gone, so a bare host in `KV_REST_API_URL` gives a generic fetch error now instead of the targeted one. readme shows the full `https://` form so low risk

checked whether the write token gets baked into the edge bundle since this runs in middleware - it doesn't, `process.env` is preserved for runtime lookup. no issue, just noting it so nobody else goes looking. the build-writes / runtime-reads split is nice

edge-middleware/redirects-upstash/tsconfig.json
```diff
@@ -2,7 +2,11 @@
   "compilerOptions": {
     "baseUrl": ".",
     "target": "es5",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": [
```
kldavis4, line 5: this reformat isn't prettier - `--check` passes on this file on main and fails now. one-key-per-line json is editor formatting, and the repo runs prettier over `**/*.json` via lint-staged

`jsx: react-jsx` and the `next-env.d.ts` change are real next 16 output though, a fresh build produces no further churn, so those should stay

`cdn/mintlify-docs-rewrite/vercel.ts` also fails the check - a `pnpm prettier` over both would clear it

PR #1558, opened 2026-08-14
[services] Next.js + Python Queues

ricardo-agz:
### Description

Adds an example that shows how to use queues across runtimes, e.g. enqueueing from Next.js and consuming from Python, or enqueuing from Python and consuming from TS

### Demo URL

https://next-python-vercel-queues.labs.vercel.dev/

### Type of Change

- [x] New Example

### New Example Checklist


services/cross-runtime-queues/backend/store.py
```diff
@@ -0,0 +1,102 @@
...
+import sqlite3
+import time
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Any, Iterator
+
+from vercel.cache import AsyncRuntimeCache
+
+RESULT_TTL_SECONDS = 3600
+LOCAL_DATABASE_PATH = Path(__file__).with_name(".queue-results.sqlite3")
```
dnwpark, line 13: We really ought to support runtime cache in `vc dev` instead of this

services/cross-runtime-queues/frontend/src/app/api/next/next-to-python/route.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  await putTask(task)
+  try {
+    const { messageId } = await send('demo-next-to-python', {
+      ...payload,
+      taskId,
+    })
+    task.messageId = messageId
+    await putTask(task)
```
dnwpark, line 20: why `putTask` twice?

dnwpark (approved):
Kinda confusing to follow all the logic. Is it worth putting the store/task tracking in a separate service?

PR #1579, opened 2026-09-21
[configcat] disable marketplace sync for retired demo

molebox:
### Description

- The hosted ConfigCat demo returns 404. Remove its advertised URL and point readers to the current ConfigCat integration guide.
- Set `marketplace: false` so future README syncs skip this template instead of updating or republishing its marketplace entry.
- This is the repository half of the retirement. The existing marketplace entry must be unpublished separately, with its public URLs handled by the publishing owner.

### Type of Change

- [ ] New Example

### Validation

- Confirmed the replacement documentation URL serves the ConfigCat setup guide.
- Confirmed the README contains no references to the dead demo hostname.
- Inspected the existing template parser: boolean `marketplace: false` returns before marketplace entry lookup/update.


tim123abc (approved):
Just to clarify the owning team is removing this from contentful?

